PR #5, opened 2017-08-09
bugfixing get_order_tsp function, returned wrong order for e.g. (0, 3…

alfrisch:
…, 1, 2)


alfrisch:
fixing a minor bug, which returned the wrong solution for TSP

antoniomezzacapo:
Thanks for pointing out the bug.

PR #23, opened 2017-09-05
Added list of gates and their matrices

rraymondhp:
Hi,

As a reference when designing a quantum circuit, I have written a tutorial that lists all gates available.

### DIFF
--- a/1_introduction/quantum_gates_and_linear_algebra.ipynb
+++ b/1_introduction/quantum_gates_and_linear_algebra.ipynb
@@ -28,13 +28,13 @@
    "source": [
     "## Quantum Gates \n",
     "\n",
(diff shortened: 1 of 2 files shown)


jaygambetta:
@rraymondhp can we remove the location for the qiskit-skd-py as for the tutorials we should assume that the used installed with pip. @ewinston is working on the readme to update this. Otherwise the rest of the tutorial looks good.

rraymondhp:
@jaygambetta  Thanks! Commenting out the location of the qiskit-sdk-py and add notice for ones who has not installed with pip.

rraymondhp:
Please review this submission  that try to list all gates that can be used from the QuantumProgram.

PR #25, opened 2017-09-09
Vaidman Detection Test

sabreitweiser:
Hi all,

This example is based one we prepared during a demonstration from Dr. Corcoles and Dr. McClure at a summer school this past June. It's a simple demonstration of interaction free measurement, based on the Vaidman bomb testing circuit, which Dr. Corcoles suggested. I've updated it to reflect changes in the API structure and added some explanatory text/figures to match the format of other examples in the tutorial.

//Alex


rraymondhp:
Hi,
I think this is a cool tutorial that exploits the difference in probability computation: quantum computing uses probability amplitudes that when used with intermediate measurement can result in quadratic difference. 

I have some details comment:
- The rotation R_y(a) is actually a rotation by a/2. That is R_y(a) |0> results in cos(a/2) |0> + sin(a/2) |1>. Would you mind fixing this? Also, when setting epsilon = pi/nsteps, after nsteps of iteration, for the case dud, it will be flipped by pi/2, which is sin(pi/2) = 1 instead of flipping by pi. 

- you can use numpy.pi instead of 3.14159 (QASM also defines pi this way)

jaygambetta:
@rraymondhp the rotation angle is defined correctly using our convention.

rraymondhp:
@jaygambetta  But, according to qelib.inc1, R_y(theta) |0> = cos(theta/2) |0> + sin(theta/2) |1> ? 
See Eq.2 of the qasm2.pdf here. That means, strictly speaking, after a rotation R_y(epsilon), the state will be approximately |0> + epsilon/2 |1> instead of |0> + epsilon |1> .
https://github.com/QISKit/openqasm/blob/master/spec/qasm2.pdf

jaygambetta:
I am not disagreeing with the math. I am disagreeing with the interpretation. If you map this to a probability or a Bloch sphere then putting a pi rotation gives you the excited state [sin(pi/2) = 1]. It is standard convention to call a pi pulse a pulse that makes 1 and a pi/2 that makes 0+1. This is a common (tiny) mistake often made and it depends if your definitions. We are defining a rotation by theta this way.

rraymondhp:
Thanks for the clarification. My comment is on the math, and I have corrected/omitted the rotation term.

sabreitweiser:
I've updated the pull request to remove the custom ket command and correctly use numpy.pi. While I'm not intimately familiar with the angle definitions, I believe this is correct based on some testing.

rraymondhp:
@sabreitweiser  thanks for the swift response. I have summarized the matrices for gates available at QASM at the tutorial below. For small x, sin(x) approx x, and thus my comment.  
https://nbviewer.jupyter.org/github/QISKit/qiskit-tutorial/blob/master/1_introduction/quantum_gates_and_linear_algebra.ipynb

rraymondhp:
@sabreitweiser  Also, please also update index.ipynb to link your tutorial and add your name.

sabreitweiser:
Ah, I see now - the rotation is correct but my approximation is wrong. I've updated it accordingly, as well as adding the example and my name to index.ipynb.

PR #31, opened 2017-09-19
Added repetition code to section 2 and description in index

decodoku:
The file 'repetition_code.ipynb' has been added to the quantum information section. A directory with data that can be analyzed with the notebook was also added. A description was added to 'index.ipynb'.


decodoku:
Hi @rraymondhp. Sounds fine to me. I have deleted the files, but I've also made sure that the ipynb has some ready made graphs. I removed reference to the data in the ipynb, and added a link to an alternative repo where it can be found.

rraymondhp:
Thank you very much for the quick response.

attp:
Hi @decodoku, great tutorial, and associated paper! Just a couple of minor comments:

- It looks like there's been a change to the standard header to the tutorials to "_The latest version of this notebook is available on https://github.com/QISKit/qiskit-tutorial._" rather than the "_For more information about how to use the IBM Q experience (QX), consult the tutorials, or check out the community._" you have there now. 
- If I were you, I wouldn't refer to the associated paper with just a "_A paper based on the results from this study can be found [here](https://arxiv.org/abs/1709.00990)._" but with it's full reference, like you have with your earlier paper. _[J. R. Wootton, D. Loss arXiv:1709.00990 (2017)](https://arxiv.org/abs/1709.00990))_ It might also be worth suggesting people read your paper for a discussion of the ready-made graphs.

decodoku:
Hi @attp, thanks for the suggestions. Done and done!

decodoku:
@smolinIBM Thanks for the suggestions.

Regarding your more serious question, I am implementing this code primarily to test out some of the techniques of QEC. Specifically, to test out the power of ancilla assisted syndrome measurements. So the ancilla measurements are there by definition, but you are right that we should check whether they are worth using at all.

I did analyze the results to see how well the stored bit can be decoded if the results from the ancillas are ignored. I did not go one step further and see how well the bit can be decoded if the CNOTs are never performed at all. I don't think there would be too much difference, but I can add the functionality to probe this question.

Since ibmqx3 is being rebooted, and will soon become ibmqx5, I will add support for the changes to device architecture. I'll commit again once that is in place, as well as implementing the changes John suggested.

PR #54, opened 2017-11-17
Added quantum emoticon notebook

decodoku:
As discussed on the QISKit slack.


rraymondhp:
I think we need to edit the `index.ipynb` file to list the hello world, but it can be done later.

PR #57, opened 2017-11-23
Fixing maxcut by correcting bit order for Pauli matrix

alfrisch:
the maxcut tutorial with an asymmetric graph gives an answer with the wrong bit order
e.g. running with
elist=[(0,1,8.0),(1,2,1.0),(2,0,2.0)]
has correct solution [1,0,0] or [0,1,1] but the result is [0,0,1] or [1,1,0]
inverting the bit order when building the list of Paulis fixes this

tested by running the notebook for various asymmetric maxcut problems


rraymondhp:
@alfrisch  Thanks for correcting this. When you change the order of the bits, I wonder if the optimization results will change too. Did you observe any change of the optimization results?

alfrisch:
The above mentioned example for maxcut on 3 nodes has the brute force solution
case = [0, 0, 0] cost = 0.0
case = [1, 0, 0] cost = 10.0
case = [0, 1, 0] cost = 9.0
case = [1, 1, 0] cost = 3.0
case = [0, 0, 1] cost = 3.0
case = [1, 0, 1] cost = 9.0
case = [0, 1, 1] cost = 10.0
case = [1, 1, 1] cost = 0.0

Running QAOA with shots=1 for ideal simulation (exact cost is -9) I get the following results with old bit order:
Final objective function is: -8.87405787925
{'001': 4164, '110': 682, '101': 152, '011': 1, '111': 1}

and for the new bit order:
Final objective function is:  -8.65769662821
{'010': 500, '100': 3831, '101': 495, '011': 174}

There is no qualitative difference between these results, still within the noise of the SPSA optimization.

PR #68, opened 2017-12-25
Added quantum emoticons and iterative phase estimation

rraymondhp:
Added two tutorials to stable: quantum_emoticons and iterative_phase_estimation


ajavadia:
@rraymondhp: can you please fix the conflict in index.ipynb?

rraymondhp:
@ajavadia  Thanks! I have fixed the small conflict.

rraymondhp:
To my understanding, if we have tutorials that work with the stable version of the sdk, we should also have them in the stable branch. We should be able to periodically update and do this, say one in two weeks.

PR #70, opened 2017-12-26
Added Bernstein-Vazirani algorithm to the stable branch

rraymondhp:
A simple and fundamental quantum algorithm experimented with oracle created without CNOT gates.


attp:
Looks good to me.

Issue #74, opened 2017-12-30
Suggestion on managing API and URL via environment vars for real devices


EvanKepner:
Hey just started this tutorial, love that you are putting this together. One suggestion for the introduction when it comes to the real device configuration - have the users put their API key and the URL into environment variables instead of into local Python files. This removes the risk of people accidentally committing to public repos and exposing their keys. 

The following would be changed:

1. README - in section 4 instruct the user to setup an `IBMAPI` and `IBMURL` environment variable e.g. in their `.bash_profile` or similar using `export` statements.
2. In the Introduction - Getting Started notebook change the imports to remove `Qconfig` and instead `import os`.
3. In the Introduction - Getting Started notebook change the "Real Devices" section to get the environment variables instead e.g. 

```python
qp.set_api(os.getenv("IBMAPI"), os.getenv("IBMURL"))
```

rraymondhp:
Hi, many thanks! This is a very nice suggestion. I wonder how to do a similar thing with Windows environment. I think this should also be considered at `qiskit-sdk-py`.

EvanKepner:
For sure! I think it varies by Windows version (unfortunately I don't have a system to test on) where you can set environment variables under the advanced system settings. 

If you are open to accepting pull requests I would be happy to add instructions in the README to provide the alternative example with `os.getenv()`. If the `QConfig.py` file is used in multiple notebooks it could take some time to update them all, but that way the audience is aware of the option if they want to make changes in their own copies of the notebooks.

rraymondhp:
Many thanks! Please feel free to add instructions in the README. I personally agree to have more options and make things simpler if possible. I think your suggestion is a good way towards, but we need to discuss it with those in `qiskit-sdk-py` too. 

But please understand that it will be one of the options at the moment because we think having all in python thru the `QConfig.py` is essential.

EvanKepner:
Excellent, I'll put in a pull request for review. Thanks! If keeping the`QConfig.py` is essential I'll move the suggestion to changing that file instead of the notebook e.g. 

```python
APItoken = 'my token from the Quantum Experience'
config = {'url': 'https://quantumexperience.ng.bluemix.net/api'}

if 'APItoken' not in locals():
    raise Exception('Please set up your access token. See Qconfig.py.')
```

Would now optionally become:

```python
import os

APItoken = os.getenv("IBMQE_API")  # if you have set this via an environment variable
[...]

PR #75, opened 2017-12-31
Add environment variable import for API token and example Qconfig to stable

EvanKepner:
Hi - in reference to the discussion on #74 I've added the following changes to the stable branch for review:

- Added alternative configuration to the README explaining how to use `os` for importing an environment variable from the shell session.
- Added an `example_Qconfig.py` to the introduction folder that can be customized. It will raise an exception if the user does not set an environment variable or has an unset config dictionary.
- Added my name alphabetically to the contributors list.

Thanks for considering this!

### DIFF
--- a/1_introduction/example_Qconfig.py
+++ b/1_introduction/example_Qconfig.py
@@ -27,9 +27,3 @@ config = {"url": "https://quantumexperience.ng.bluemix.net/api"}
 
 if APItoken is None:
     raise Exception("Please set up your access token. See Qconfig.py.")
-
-if config is None:
-    raise Exception("Please set up your config parameters. See Qconfig.py")
-
-if __name__ == "__main__":
-    pass
--- a/1_introduction/example_Qconfig.py
+++ b/1_introduction/example_Qconfig.py
@@ -1,0 +1,35 @@
+"""An example Qconfig file you can customize.
+
+Set your configuration parameters to be imported into the notebooks.
+
+The APIToken can be set as an environment variable in your shell session. For
+example, in your ~/.bash_profile or ~/.bashrc add:
+
+.. code-block:: bash
+
+    export IBMQE_API="your_secret_api_string"
+
+and restart your terminal session. This will add the ``$IBMQE_API`` as an
+environment variable that can be accessed by the script.
+
+Once imported, you can check the values have set using print:
+
+.. code-block:: python
+
+    import Qconfig
+    print(Qconfig.APItoken, Qconfig.config['url'])
+
+"""
+import os
+
+APItoken = os.getenv("IBMQE_API")  # change to match your variable name if needed
+config = {"url": "https://quantumexperience.ng.bluemix.net/api"}
+
+if APItoken is None:
+    raise Exception("Please set up your access token. See Qconfig.py.")
+
+if config is None:
+    raise Exception("Please set up your config parameters. See Qconfig.py")
+
+if __name__ == "__main__":
+    pass
--- a/README.rst
+++ b/README.rst
@@ -64,6 +64,22 @@ Briefly, the steps are:
             if 'APItoken' not in locals():
                 raise Exception('Please set up your access token. See Qconfig.py.')
 
+    4. Alternatively, you can set your API token in an environment variable for your
+       shell session and import it into the ``Qconfig.py`` file using ``os``.
+       For example:
+
+       .. code:: python
+
+           import os
+           APItoken = os.getenv('your_API_token_environment_variable')
+           config = {'url': 'https://quantumexperience.ng.bluemix.net/api'}
+
+           if APItoken is None:
+               raise Exception('Please set up your access token. See Qconfig.py.')
+
+        An example ``Qconfig.py`` file has been included in the introduction
+        notebook folder.
+
 2. Install `Jupyter <http://jupyter.readthedocs.io/en/latest/install.html>`__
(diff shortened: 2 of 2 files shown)


rraymondhp:
Hi,
happy new year! 

Thank you for the suggestion. I think we better not modify the `Qconfig.py`. One way of doing this is to check for the environment variables. If we can find the key, then we do not import `Qconfig.py`. Otherwise, we proceed to import it and proceed as usual. This way, we can make sure we still can work with those using the `Qconfig.py` and the environment variables. 

Perhaps something like:
```
APItoken = os.getenv("IBMQE_API")  # change to match your variable name if needed
config = {"url": "https://quantumexperience.ng.bluemix.net/api"}

if APItoken is None:
    import Qconfig
    APItoken = Qconfig.APItoken
```

EvanKepner:
Happy New Year! Maybe I misunderstood the intention, can you help clarify this for me? 

When I look through the tutorials using GitHub Search I only see Qconfig used as an import for the API Token and the Config url. In the [SDK docs the Qconfig files example](https://www.qiskit.org/documentation/install.html?highlight=qconfig#configure-your-api-token-and-qe-credentials) has more values in the config but same core structure. I saw you have `Qconfig.py` in the `.gitignore` for the tutorials as well to keep it from syncing to GitHub. This works if I've forked the repo and gotten the `.gitignore`, or know to include it on my own as an ignored file, but if I've started a new repo I might want the Qconfig file to be included (or just forget). 

The modification I'm suggesting for the `Qconfig.py` file allows the users to keep their Qconfig in their repo without any sensitive data like the API Token. The change to check `APIToken is None` is because `os.getenv()` will return `None` if the environment variable isn't set, so that variable name will still be in `locals()` and the prior check would pass. Modifying `Qconfig.py` directly means that all of your tutorial notebooks would continue to work without modification to make use of the environment variable.

In the above example: importing `Qconfig` and setting the `Qconfig.APItoken` value if the environment variable is not present, implies that I've got the API Token in plaintext in the local directory somewhere, which doesn't address the concern. 

If there's a concern with the modification in code maybe we're better off with just text e.g. "as an alternative, you can set your API Token as an environment variable and import it with `os`." and also add a warning to the user to add "Qconfig.py" to their `.gitignore`. If you do a GitHub Search for "Qconfig apitoken" you'll find a few examples of publicly viewable API keys in code. There is a [sensitive data cleaning mechanism](https://help.github.com/articles/removing-sensitive-data-from-a-repository/), and the user could always regenerate the key, but it's better to avoid this in the first place.

Let me know what you think and I'll be happy to make any modifications. Thanks again for considering this!

rraymondhp:
Hi, thank you very much!
 
My intention is to keep the tutorial compatible with the `qiskit-sdk-py`. I want to make sure that the same `Qconfig.py` file used at the sdk also works for the tutorial. This is because the tutorial is a derivative of the sdk. We better have the tutorial support the same `Qconfig.py` as the sdk. 

But having a capability to get the credential token from the environment without having it from the `Qconfig.py` is a nice feature. I think it will be good to post that as an issue at the sdk too.

EvanKepner:
Got it, that makes total sense, thanks! I'll take your suggestion and open an issue related to this on the SDK repo directly. I'll update this PR once that discussion happens (or if you'd rather close it for organizational reasons I can open a new one later).

rraymondhp:
Thanks! I will note your idea in the wiki for reference. Until that, let's keep this issue open.

rraymondhp:
Hi @EvanKepner 
My colleague proposed the following template of `Qconfig.py`. what do you think? 
```
# Before you can use the jobs API, you need to set up an access token.
# Log in to the Quantum Experience. Under "Account", generate a personal
# access token. Replace "None" below with the quoted token string
# or set an environemnt variable "IBMQX_TOKEN".

import os

APItoken = os.getenv('IBMQX_TOKEN', None)
if not APItoken:
  raise Exception("Please set up your access token. See Qconfig.py.")

config = {
  "url": 'https://quantumexperience.ng.bluemix.net/api'
}
```

1_introduction/example_Qconfig.py
```diff
@@ -0,0 +1,35 @@
...
+"""
+import os
+
+APItoken = os.getenv("IBMQE_API")  # change to match your variable name if needed
+config = {"url": "https://quantumexperience.ng.bluemix.net/api"}
+
+if APItoken is None:
+    raise Exception("Please set up your access token. See Qconfig.py.")
+
+if config is None:
```
t-imamichi: I think this is redundant because `config` is set above.

1_introduction/example_Qconfig.py
```diff
@@ -0,0 +1,35 @@
...
+    raise Exception("Please set up your config parameters. See Qconfig.py")
+
+if __name__ == "__main__":
```
t-imamichi: It seems redundant because Qconfig.py is not a stand-alone script.

1_introduction/example_Qconfig.py
rraymondhp: Thanks. 
@EvanKepner  Once you delete the lines 28 - 31, I will merge your PR. 
Thank you very much for the improvement!

EvanKepner:
Thanks @rraymondhp! The changes are made!

I had included the assertion for `config` just so it was clear it was a required property when modifying the example. The `if __name__` lines are an idiom to ensure that if the script was run as a standalone for some reason it would simply `pass` unless specified. Habit of mine to keep in place but doesn't hurt it to be removed.

PR #79, opened 2018-01-07
Introduction notebook for composer examples in beginners guide

EvanKepner:
Hey - I went through the beginners guide and found it really useful. One thing I had wished for was the equivalent SDK code for each composer example and I couldn't find it. I created this notebook as a companion example for the different composer diagrams. Thanks for your consideration!


ajavadia:
Thanks for this @EvanKepner! A few comments:

* Can you please include a link to your notebook in the index.ipynb (under section 1)?
* Can you include all images in the `images/` directory of root, and delete the `1_introduction/img` directory? (Make sure all image references in the notebook get updated).
* Please remove 1_introduction/Qconfig. This is already included in the QISKit SDK repository.

EvanKepner:
No problem! I think I'll have to close this PR and open a new one to make the branch update work. I misunderstood stable vs. master and thought it was syncing to features of the SDK specifically. 

I included the `example_Qconfig.py` in the intro notebooks folder so that if people renamed it and re-ran the notebooks directly it would import (since all the notebooks have `import Qconfig`). If It was up a level then there would be the additional step of copying it down. My assumption was that having it in the intro was more clear as a starting place for people to use it in the future.

Once I put in the new PR against master I will mark this one as closed.

EvanKepner:
Marking as closed since this is replaced by https://github.com/QISKit/qiskit-tutorial/pull/81

rraymondhp:
@ajavadia  Thanks!
I think the example of `Qconfig.py` provided by @EvanKepner  is better than the one in the sdk. Moreover, users still need to copy (or, soft-link) the `Qconfig.py` of the SDK to the root directory of the tutorial to make thinks work. For this reason, I like Evan's proposal.

PR #89, opened 2018-01-31
fix relaxation notebook with barriers

ewinston:
Compilation for the backend was causing the identities used in the relaxation_and_decoherence notebook to be optimized out. This change puts barriers between the identities operations to prevent that.


jaygambetta:
@ewinston this has errors.

Issue #92, opened 2018-02-03
reworking the tutorial so that it is easier to manage


jaygambetta:
We are going to split the tutorial into three sections. 

We are going to add to a WIP PR that i will set up and if you want to help in the rework that would be great. Either you can help by commenting here or helping in the content. The names are the assigned people that are going to coordinate the different sections. 

# sections

* **Hello World**: Since quantum computing is so new, we want to find the best *Hello Quantum World* program and welcome submissions here. 
* **Reference**: These notebooks demonstrate how to use QISKit and explore quantum information science, acting as a reference book for QISKit. They will be kept up to date with QISKit SDK updates. They are organised into various topics:
  1. Getting started with QISKit  (me)
    - Getting started
    - Working with backends (including compiling and running)
    - Using different gates
    - Visualisation of quantum states
  2. Introduction to quantum information science  @attp 
    - Superposition and entanglement
    - Entanglement revisited
    - Quantum teleportation and superdense coding
  3. Understanding your quantum computer @dtmcclure 
    - Relaxation and decoherence
    - Quantum tomography
    - Random Benchmarking # to be written
  4. Working an approximate quantum quantum computers  @adcorcol 
    - Variational Quantum Eigensolver
    - Small error correcting codes # to be written
    - Error mitigation # to be written
  5. Examples of quantum algorithms @rraymondhp 
    - Phase estimation
    - Deutsch–Jozsa algorithm
    - Bernstein-Vazirani algorithm
    - Simon's algorithm
    - Grover's algorithm
    - Shor's algorithm
  6. Having fun with quantum computers 
    - Battleships
    - 
    - 
* **Appendix** This is where the rest of the tutorials are. They are not guaranteed to work with the latest version of the QISKit SDK. They are organised into various topics:
  1. Advanced QISKit features
  2. More on quantum information
  3. Further quantum algorithms and applications
  4. Everything else

jaygambetta:
To help out please commit to https://github.com/jaygambetta/qiskit-tutorial.

rraymondhp:
@jaygambetta  Would you mind updating the getting started QISKit tutorials? 

@dtmcclure  Do you think you can complete the random benchmarking within this week? 

@adcorcol  Do you think you can complete the rest of tutorials for approximate quantum computers

dtmcclure:
[...]

PR #93, opened 2018-02-03
[WIP] tutorial rework

jaygambetta:
Start of the WIP tutorial rework. #92


rraymondhp:
Jay,
Thank you for fixing the readme and install. They are much simpler and easier now.

jaygambetta:
im working in the index now and then i have my section done by end of day. 

One question is if you guys agree we can also remove the subsections from the readme and as you said there could be a readme in each of the folders.

rraymondhp:
I agree to remove subsections from readme (they can go to index). Also, readme.md files are those viewed from github viewer (browser). IMHO, they should contain concise intro and links to installation, index, contributors, and other non-technical things.

jaygambetta:
OK i agree and i make this change. Did not want to change it too much from what you guys had. But this is what i want with the readme as well

jaygambetta:
closing and doing again using git mv x etc so that history of files remains.

PR #94, opened 2018-02-06
Entangler map for LiH

nikomoll:
The original entangler map is only sufficient for H2 (two qubits)


ajavadia:
@nikomoll Thanks! The entangler map is indeed missing for LiH.
But it is better to get that from the backend itself. Each real device has different connectivity, and simulators have 'all-to-all' connectivity.

Can you please make this change? Something like:
```
entangler_map = Q_program.get_backend_configuration(device)['coupling_map']
if entangler_map == 'all-to-all':
    entangler_map = {i: [j for j in range(n) if j != i] for i in range(n)}
```

nikomoll:
@ajavadia   I added your suggestion in the new commit.

ajavadia:
This looks good to merge. I don't have write access though.

PR #118, opened 2018-04-14
tutorial rework

jaygambetta:
Adding the new tutorial format in


jneer:
With all this reorganization, there might be several broken links in various places. At least there's the "Tutorials" button on the qiskit.org front page.

Another quick comment: In the Deutsch-Jozsa tutorial, the final cell should be run at a time when ibmqx5 is available ;)

PR #138, opened 2018-04-16
Confirmed the tutorial works, and modified printing the version

rraymondhp:
I have confirmed the Bernstein-Vazirani tutorial works for simulators and real devices.
I also added codes to correctly identify installed the versions of important packages, such as, qiskit, IBMQuantumExperience, etc. 

1. We have to first query the available backends (so that we are connected to remote devices)

2. The versions of important packages are queried with the commands below listed at the bottom of the tutorial (the default `from utils import version; version.version_information()` can give wrong results as it is only listing the content of `requirements.txt`)
```
!python --version
print()
!pip show qiskit
print()
!pip show IBMQuantumExperience
print()
!pip show numpy
print()
!pip show scipy
print()
!pip show matplotlib
```


jaygambetta:
@rraymondhp could we also use a function like we did in the superposition for getting the best device so that the backend is not hard coded in. Also i agree with @attp we should try to remove the warnings.

rraymondhp:
The warning is from deprecated function, I am not sure how to remove them. 
For the available backends, I am using 15 qubits or more. Do we have choices?

rraymondhp:
@jaygambetta  IMHO, the selection of real devices can be done inside the execution (from qiskit-sdk-py) without users defining the function, like what you wrote, to select real devices. That way the program can be much simpler. If you agree I will make a PR to the qiskit-sdk-py.

rraymondhp:
@jaygambetta  @attp  I have revised the program. There are no warnings anymore.

jaygambetta:
This is great. 

However can you remove you version addition and just put the old way for now in this tutorial (so we have  all tutorials in sync). Then in a new pr add this version notebook so that we can check on all operating systems. I will get @diego-plan9 review it. Then i think we can do a third pr where we go add it too all pr so that they have the correct version they were run on.

rraymondhp:
Thanks. I am modifying the "version.ipynb" so that all tutorials that run them should be okay. And I can confirm that they are OS independent. I think we need these so that users do not get confused with the version installed in the system.

rraymondhp:
@jaygambetta  This is the current output of version checker. It maintains the old output (required package) while also displaying what actually installed.
image

rraymondhp:
Otherwise, the contents of the `requirements.txt` must be updated to the newest one. The current one is old and wrong. For example, the content below is old version (v0.4.10) and this is always shown regardless of what is truly installed. 

https://github.com/QISKit/qiskit-tutorial/blame/master/requirements.txt

rraymondhp:
I have confirmed it works and merged the PR. The version checker will be adjusted later.

PR #146, opened 2018-04-21
robust Qconfig template; fixed bug in retrieving Qconfig template 

sathayen:
- Robust  Qconfig template - updated `Qconfig.py.template` - you can either hardcode the APItoken directly in `Qconfig.py` OR set it as an environment variable. If the token is hard coded in Qconfig file it OVERRIDES the value of the environment variable. There is also a new convenience method `update_token` to update to the current value API token. 
- Fixed a bug in retrieving the template in DSX.ipynb

### DIFF
--- a/DSX.ipynb
+++ b/DSX.ipynb
@@ -45,7 +45,9 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {},
+   "metadata": {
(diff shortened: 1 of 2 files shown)


rraymondhp:
Thanks. I confirmed that it is working when no Qconfig.py exists in user's folder.

PR #147, opened 2018-04-24
Qcvv apr22

dtmcclure:
Fixing QCVV notebooks as discussed. Details in individual commit messages.


jaygambetta:
@diego-plan9 can you have a look as i see that the version notebook is changed in this merge request. I know that you are working on a new way of doing this.

rraymondhp:
@attp  @jaygambetta  Thanks for reviewing.  @diego-plan9  and I agreed to remove version checking and add explanations about what version should be used. I am going to do that once we have all notebooks in place. It should be easy by just replacing the `version.ipynb`.

PR #149, opened 2018-04-28
Adding quantum walk tutorial to the Appendix

rraymondhp:
This is a tutorial written by an intern in IBM Research - Tokyo.


attp:
Please make sure that all the maths notation renders properly in GitHub.

rraymondhp:
@attp  Thanks. Many complicated formulas and rewriting all of them to adjust with github rendering is difficult. Moreover, because we will use jupyter notebook, I think we should not be too relying to github. 
You can use nbviewer rendering to see better like here: 

https://nbviewer.jupyter.org/github/rraymondhp/qiskit-tutorial/blob/4a311478d3bf98f0be6fb0ae536c0b400e47948c/appendix/algo_app/topological_quantum_walk.ipynb

PR #150, opened 2018-04-30
Added the Simon Algorithm

rraymondhp:
I cannot run the algorithm on real devices due to timeout error, but I am sure the algorithm is correct (as confirmed in the simulator)


attp:
I think you should spend some more time describing Simon's problem - perhaps giving an example with a 3-bit problem, and comparing the classical versus quantum complexity, rather than just stating them.

jaygambetta:
but please rebase and fix up the conflicts and edit for 0.5 before we merge

rraymondhp:
I have rebased, fixed the conflict, and edited the command following v0.5

Issue #144, opened 2018-04-18
Provide a way for checking the tutorials when a new qiskit version is released


diego-plan9:
## Description

After some discussion spawned from #135 and the tutorials as a whole, it seems that having an easy, ideally single-shot way of checking if the tutorials are working against a new version of `qiskit` (or other dependencies in general) would facilitate keeping them in sync.

As a simple approach, a command that can be executed locally by a tutorial repository maintainer and clearly outputs which ones are working and which ones are not seems a good compromise. We can make use of the Python test framework for paving the way towards having a more integrated system in the future if needed.

## Your Environment

* QISKit version:
* Python version:

rraymondhp:
This has been solved thanks to @diego-plan9

PR #151, opened 2018-04-30
Add helper for running the tutorials as tests

diego-plan9:
Add `utils/test/test_tutorials` for being able to run the notebooks as
unit tests manually using the standard Python unittest framework. The
implementation automatically discovers the "*.ipynb" files under
`reference/`, and attempts to run the Python code in each file as a
single unit test.

Some instructions on how to use it can be found in the docstring at `utils/test/test_tutorials.py`, but the goal was to keep it as close as regular tests as possible, ie:
```
python -m unittest
```

Solves #144 .

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 Qconfig.py
 
 __pycache__/
+.cache/
 
 .ipynb_checkpoints/*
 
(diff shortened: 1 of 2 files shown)


rraymondhp:
Thank you very much! As we have tutorials that can be used for testing, I am merging it.

Issue #155, opened 2018-05-12
rewriting the tutorials for version 0.5


jaygambetta:
## Your Environment

* QISKit version: 0.5
* Python version: 3.6

jaygambetta:
@attp and @rraymondhp which ones are left to rewrite

jaygambetta:
closing as focusing on just the tutorials that are remaining.

PR #158, opened 2018-05-14
Editing QIS tutorials to be compatible with QISKit 0.5.0

attp:
This addresses Issue #155 for the quantum information science tutorials.


decodoku:
There are still some `Q_program.execute` calls in entanglement_revisited.ipynb. Is this intentional?

attp:
@quantumjim and @rraymondhp, I've now removed the execute calls from all the QIS tutorials.

PR #161, opened 2018-05-16
Edited Battleships to be compatible with qiskit 0.5.0

quantumjim:
This addresses Issue #155 for the battleships in the games folder.

An additional change is to spin out the exclusively classical parts into a separate .py file. This gives the notebook a focus on the quantum, and the .py can also be used for further battleships variants.


quantumjim:
I have been doing this on and off for a while, and it seems that @abbycross did some proof-reading in that time. I have now manually implemented Abby's changes in my version. So though git still sees a conflict, I think it is effectively resolved.

rraymondhp:
@quantumjim  Thanks! Let me see if I can resolve the conflict.

quantumjim:
It seems there were a few more changes to be made, which I did. The conflict is now resolved.

rraymondhp:
Thanks! Just a small request: can you remove the version information at the bottom of the notebook. We do not need that anymore. I am going to remove the version from all notebooks.

quantumjim:
This notebook will no longer work due to the recent changes to the remote simulators.

No simulators (local or remote) are currently working for me, so I am not yet able to create a new working version.

jaygambetta:
@quantumjim  @atilag is fixing the remote now. He will put together an qiskit update when he is done can you let us know if there is still errors in the remote. Sorry for the inconvenience and the names should be in steady state now for the remote. 

For the local what error are you getting?

quantumjim:
It's now back to a working version, so all is well once more.

@jaygambetta My local simulator issue is the same one someone else flagged up on the slack
https://qiskit.slack.com/archives/C7SJ0PJ5A/p1526501921000448
so @atilag is on it :)

PR #169, opened 2018-05-24
tested the audio environment

rraymondhp:
Adding requirement and instruction to install audio-related libraries to run the notebook.


rraymondhp:
This is a fix to remove the wav files (temporary) until I can confirm the legal status of the files.

PR #173, opened 2018-05-25
Audio File Updates

APerruzziIBM:
I have created my own versions of the files, and uploaded those in place of the old ones. These files were derived directly from Vocabulary.com.

The tutorial notebook now contains the citation, with a link back to the original site.

The README in the audio folder also contains the citation, with a link back to the original site.


rraymondhp:
Hi @APerruzziIBM  
Can you check and fix the conflict?

APerruzziIBM:
@rraymondhp, I've attempted to resolve them offline (as is apparently required because they're complex), but this isn't a branch, so 'git status' on the local repository isn't returning anything. This is my first time resolving conflicts that are too complex for the browser, so there may be something I'm missing here. Let me know what I'm doing wrong, or if it makes sense to create a new pull request starting from an updated master branch.

PR #175, opened 2018-05-26
edited ipea to use v0.5 api

rraymondhp:
Iterative phase estimation using api of v0.5.0


rraymondhp:
this is a hoftix to use api of v0.5

PR #204, opened 2018-06-06
Adding getpass method of API token input to all reference tutorials.

attp:


attp:
Will open a new request with the merge conflicts.

PR #211, opened 2018-06-08
Fix function name in quantum_gates_and_linear_algebra

ikkoham:
- Replace norm with fidelity
- Add description of fidelity

### DIFF
--- a/reference/tools/quantum_gates_and_linear_algebra.ipynb
+++ b/reference/tools/quantum_gates_and_linear_algebra.ipynb
@@ -2450,6 +2450,18 @@
    ]
   },
   {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "[Fidelity](https://en.wikipedia.org/wiki/Fidelity_of_quantum_states) is useful to check whether two states are same or not.\n",
+    "For quantum states $|\\psi_1\\rangle$ and $|\\psi_2\\rangle$, the fidelity is defined by\n",
+    "$$\n",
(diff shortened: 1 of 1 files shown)


jaygambetta:
i would rather not call this fidelity. As fidelity we should save for mixed states.

ikkoham:
Thank you. Fidelity for pure state is the special case of the fidelity for mixed state, so these definitions are consistent.
But if you say so, I will change my proposal.

Since this is not norm, the name should be changed.
How about `state_overlap`, I found in [another notebook](https://github.com/QISKit/qiskit-tutorial/blob/master/reference/qis/superposition_and_entanglement.ipynb)?

jaygambetta:
sure

rraymondhp:
I see. So, the `fidelity()` should take two mixed states (matrices) as inputs. The returned values will be the same if the input matrices are of rank one (or pure states).

ikkoham:
I just found an appropriate function to calculate the fidelity.
```from qiskit.tools.qi.qi import state_fidelity```
This `state_fidelity` is compatible with the fidelity for pure states.
This `state_fidelity(desired_vector,qc_state)` runs well:smile:
So I think using `state_fidelity` is better if this is a standard tool.

ikkoham:
DRY. so i use the `state_fidelity` function which has already been defined.

> By the way, I suggest to have a separate notebook to explain the concept of fidelity in more details.

I agree. But this is a future task.

ikkoham:
Thanks!

PR #216, opened 2018-06-19
Update W State 3 - Monty Hall Problem Solver.ipynb

PierreDC:
Only one of the three successive equations for C3 was rendered (Why? I don't know). I verified that they all three appear in this corrected version. Also two small typos corrected.


PierreDC:
Thank you to all the reviewers. This was the first pull request!

PR #222, opened 2018-06-20
state_distribution_in_qubit_chain: images directory, equations fix and links fixs and updates

qubeat:
Resolving problem with images and broken links. Images now are in separate directory (and svg format is converted into png).  "~" is not used now in equations. In such a case the notebooks are properly rendered in GitHub and other web viewers.

One link to 3D model fixed and two links to QISKit tutorial notebooks are updated.


rraymondhp:
Thank you...

qubeat:
Many thanks to reviewers! 
I included only corrections (no planed minor updates), because found that version is slightly modified in comparison with submitted and so additional updates could produce more complex diffs. Maybe more later, it looks not so essential as rendering problems.

rraymondhp:
You're welcome. Thank you very much for the great tutorial. 

If you want, you can also add a link to your github repository that contains the modified (and perhaps the newer) version of the notebook.

PR #223, opened 2018-06-20
Update W State 1 - Multi-Qubit Systems.ipynb

PierreDC:
After the submission of this project, I explored the idea that the deterministic obtaining of a state W can be summed up as the application of a single reversible gate to a set of qubits in the ground state. Hence this mathematical appendix which presents the gate in question in its matrix form for the case of three qubits. Formulas and equations have been established and verified using the SymPy Python library.

### DIFF
--- a/appendix/teach_me_qiskit_2018/w_state/W State 1 - Multi-Qubit Systems.ipynb
+++ b/appendix/teach_me_qiskit_2018/w_state/W State 1 - Multi-Qubit Systems.ipynb
@@ -615,14 +615,14 @@
     "\n",
     "For a three-qubit W state, the equation in Dirac notation is:\n",
     "\n",
-    "$$W_{3}\\otimes\\, |000 \\rangle=\\sqrt{\\frac{1}{3}}(|100\\rangle + |010\\rangle + |001\\rangle)$$\n",
+    "$$W_{3}\\;\\, |000 \\rangle=\\sqrt{\\frac{1}{3}}(|100\\rangle + |010\\rangle + |001\\rangle)$$\n",
     "\n",
     "which becomes in matrix notation:\n",
     "\n",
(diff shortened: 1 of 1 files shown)


rraymondhp:
Hi,
thanks for adding this. You may have noticed that there is a function in QISKit that allows you to prepare arbitrary state starting from all-zero state. It is detailed here:
https://nbviewer.jupyter.org/github/QISKit/qiskit-tutorial/blob/master/reference/tools/quantum_gates_and_linear_algebra.ipynb#Arbitrary-initialization

PierreDC:
@rraymonddhp : Thank you for your comment. Indeed I am aware of this arbitrary initialization presented in QISKit. It occurred to me that this would test experimentally the gate's reversibility by submitting a W state to the circuit corresponding to the gate's dagger. However we can also obtain this initial state W by the circuit described in my submission and continue with the reverse circuit. Do you think adding an experimental test of the reversibility by any of these ways would be interesting? If so, I may try.

rraymondhp:
@PierreDC  Thank you. 
In `Mathematical appendix`, you use `\otimes` to denote matrix-vector multiplication but this may be confused with tensor product. I think it is better to remove the `\otimes`. 

About the gate reversibility, I am not sure what you want to show. Assuming we have a good compiler, then if you combine the circuit which generates W state with the circuit that invert the W state into zero, the compiler will output identity (no operation). 

Independent of this, I think your description of generating W states for arbitrary number of qubits is very good, and perhaps it can be used as benchmarking the quantum computers (because it requires significant number of cnots and gates).

PierreDC:
@rraymonddhp OK I am not familiar with the commits and pull requests. In particular I didn’t find how to make corrections to a submitted pull request. Can you help me in the task of removing these confusing \otimes?
I agree with your “no operation” remark. However, this is also valid when you prepare an arbitrary initialization.
What interests me at this time is the error analysis in these circuits, but I am lacking background and experience about that. Using the W state model as a benchmarking method is indeed a promising idea.

rraymondhp:
Hi @PierreDC ,

To edit the PR is quite easy actually. Switch to the branch `patch-2` that you are going to merge with the master of this repository. Then, you edit the file you want to change. 

Once you have done the edition, you can commit the changed files and then push them to the `patch-2`. The changes will be automatically applied.

rraymondhp:
About the error analysis, it will be interesting to see what happens. 

>What interests me at this time is the error analysis in these circuits, but I am lacking background and experience about that. Using the W state model as a benchmarking method is indeed a promising idea.

PierreDC:
@rraymondhp "_About the error analysis, it will be interesting to see what happens._"
With an earlier version of this program, I could obtain an estimation of the classical fidelity  by calculating the Bhattacharyya coefficients for the measurement performed by running circuits |W2⟩ to  |W5⟩. I know that this is not the state of the art for error analysis (it was inspired by the paper by S. Deffner "Demonstrations of entanglement assisted invariance on IBM's Quantum Experience" https://arxiv.org/abs/1609.07459).

fideliy

rraymondhp:
@PierreDC  Thanks! It seems interesting to be able to see how the error goes as we create W states with more qubits.

PierreDC:
Thanks a lot to the reviewers!

PR #224, opened 2018-07-02
Tutorial of matplotlib_circuit_drawer

t-imamichi:
I wrote a tutorial about `matplotlib_circuit_drawer`, which is introduced in qiskit-terra 0.5.5.


t-imamichi:
Thanks. I added the header.

PR #226, opened 2018-07-09
Update reference tutorials to QISKit 0.5.6

attp:
Removed use of QuantumJob
Changed status == "available" to "operational"


jaygambetta:
i am going to merge so we fix it but we should also change it to use the filter.

PR #228, opened 2018-07-11
Use `least_busy` filter and `operational` in status

ajavadia:
The tutorials are updated to work with the latest Qiskit version:
- Built-in least busy filter replaces multiple manual functions for discovering `lowest_pending_jobs`
- One use case of checking for `available` is turned into checking for `operational`


rraymondhp:
Hi,
This seems very nice. But I wonder if selecting backend before checking the number of qubits used is good. What happens if the program uses more qubits than those available in selected backend? 

For example, in this tutorial the least busy backend was chosen before declaring the number of qubits...
https://github.com/ajavadia/qiskit-tutorial/blob/157f0b8ecbea58e4cdf487655a13d912d92a4161/hello_world/quantum_world.ipynb

ajavadia:
I didn't change the functionality of the current code (i.e. in the current code, also, the number of qubits is not checked).

It is actually possible to filter by number of qubits. For example this selects devices with more than 5 qubits:
```python
my_filter = lambda x: x.configuration['n_qubits'] > 5 and not x.configuration['simulator']
available_backends(my_filter)
```
But I don't think it is necessary, since all those examples have fewer than 5 qubits anyway. The only example that really needs a 16 qubit device is the quantum emoticon, and there the "ibmqx5" is explicitly requested. If you want, I can add this extra filter though.

rraymondhp:
Thanks for clarifying this.
I wonder if the least_busy selection can be performed with passing the circuits as arguments. The filter can know about the number of qubits from the circuits and select applicable backends wisely.

attp:
Yes, we should add that to either the Getting Started or Working with Backends tutorials.

ajavadia:
Ok I rewrote parts of the Working with Backends to introduce these filters, and also describe the concepts of provider/backend/job better.

I also edited the Getting Started to choose a device based on having enough qubits and the least busy queue.

PR #230, opened 2018-07-21
Generalized W States:  Benchmarking Algorithms by Quantum Tomography

PierreDC:
This pull request concerns a tutorial presenting the comparison of two algorithms by quantum tomography for the deterministic creation of generalized W state systems.
Both algorithms are expressed by recursive formulas resulting in compact programming of the corresponding circuits.
The first algorithm is specific for the W states, with linear progression of the number of gates. The second one is based on the arbitrary initialization of a state vector, with exponential progression. The benchmark is the quantum fidelity. Measurements were made using the ibmqx4.
The tutorial is proposed for the section:
https://github.com/Qiskit/qiskit-tutorial/tree/master/appendix/more_qis, but the reviewers may of course decide otherwise.

NB:
Small issue: when checking the final version of the notebook, there were problems when executing the part of the program calculating quantum fidelity on the real device: “502 Bad Gateway: Registered endpoint failed to handle the request.” or QISKitError: 'QISKit timed out'.  This was not constant and occurred also when executing again the older versions. Performance variation of ibmqx4? Or change in the management policy of the job queue?
Possible point to discuss: The “W state” for n = 2 is also the  Psi + Bell state, therefore accessible by another more usual algorithm. However, this does not change the fidelity which is experimentally equivalent for the three algorithms. In my opinion, mentioning this detail could be detrimental to the clarity of the tutorial.


rraymondhp:
Hi,
Thank you! I think this is great.

I have two suggestions:
1. Would you mind avoiding `QuantumProgram()` and use instead `QuantumCircuit` etc. I see you have already used `QuantumCircuit` but there are parts of your program which use `QuantumProgram()`

2. When using real-device backends, I suggest use asynchronouse job queueing. Please refer to this tutorial for how to use it:
https://nbviewer.jupyter.org/github/QISKit/qiskit-tutorial/blob/master/hello_world/quantum_world.ipynb

PierreDC:
@rraymondhp 
Hi,
Thank you for your suggestions and encouragements.
For the first point, I think that I can manage that relatively easily but I shall have to test the program again. That will be my job for tomorrow.
For the second point, do you mean introducing the loop with status check after time lapse? I don’t see  in the notebook something specifically about asynchronous job queuing.

rraymondhp:
Hi,

Yes, the second point is by looping and checking if the job has been done. This can prevent you from having timeout error.

PierreDC:
@rraymondhp 
Hi,
I cannot remove Quantum Circuit from my code concerning the tomography on a real device. For me it is obviously the same issue as this one:
https://github.com/Qiskit/qiskit-tutorial/issues/193
May I suggest continuing without this change for the moment, and when issue 193 is resolved, it will be easy to modify my code.

rraymondhp:
Hi,
Sorry I forgot about that issue. You're right about it. Please continue and once you finished with the WIP please let us know.

PierreDC:
Thank you,

Same problem with the asynchronous queueing: 
Quantum_ Circuit.execute does not accept anymore keyword arguments 'silent' and ‘wait’. 
And for the solution you indicated to me at:
https://github.com/Qiskit/qiskit-tutorial/blob/master/hello_world/quantum_world.ipynb
I have also to wait for issue 193 to be solved. 

Therefore, be so kind to consider that I am done with WIP.

PierreDC:
Thanks for requesting reviews.
I subscribed to https://github.com/Qiskit/qiskit-tutorial/issues/193
So I will be informed when this is solved and act accordingly.

Meanwhile I hope that the current version will be accepted for merging.

PR #233, opened 2018-07-23
Change QISKit to Qiskit at the files at the root directory

rraymondhp:


attp:
We should also change ACQUA to Aqua.

rraymondhp:
Thanks @attp . I wonder if @pistoia  and his team can do that...

rraymondhp:
I notice that the repository is still `acqua`...
https://github.com/Qiskit/qiskit-acqua

PR #235, opened 2018-07-23
added cvxopt

rraymondhp:
To solve this issue:
https://github.com/Qiskit/qiskit-tutorial/issues/234

### DIFF
--- a/QISKitenv.yml
+++ b/QISKitenv.yml
@@ -5,6 +5,11 @@ dependencies:
 - notebook
 - scipy
 - cvxopt
+- scikit-learn
+- cython
+- matplotlib
+- sympy
+- pillow
 - pip:
   - qiskit
   - qiskit-acqua
--- a/QISKitenv.yml
+++ b/QISKitenv.yml
@@ -4,6 +4,7 @@ dependencies:
 - pip
 - notebook
 - scipy
+- cvxopt
 - pip:
   - qiskit
   - qiskit-acqua


rraymondhp:
the latest yml is modified to handle: 
https://github.com/Qiskit/qiskit-tutorial/issues/244

After all the names have been changed, I will also change the name of yml file and documentations.

diego-plan9:
Yes, I'm not sure there is really a truly 100% best practice when it comes to dealing with a Python version transition - that's why I was suggesting "soft-recommend" Python 3.6 instead of hard-recommending it via pinning `python<=3.6`. Either way, hopefully the packages will eventually update sooner rather than later - maybe the original PR without the `python<=3.6` will suffice during that time, and the hopefully small number of issues can be dealt with individually?

ntwalibas:
I happen to deal with packages regularly so maybe my opinion on this may be of some relevance.  
The decision to restrict to Python 3.6 is indeed the correct decision.  
It is ALWAYS best to lock users to the version that is known to work for the original developer so that further fixes are made against that.   
So ideally, a version should be specified for all packages, not just Python. Failure to do so generally lands one in a dependency hell: you fix issue for one user, you create a new issue for another. Or you fix an issue with one package, only to break another package.

One more thing to make this pull request robust is to specify channels.  
You will note that in the issue #244, I am specifying the channels for each package.  
Package authors sometimes maintain multiple channels for different reasons: it can OS specific channels, channels for binaries, channels for sources, etc.

And this happens to be the case for `cvxopt`. In the present case, installation is happening from the defaults channel which requires compilation on the user's machine and this will fail for some users.  
Instead, specify the `conda-forge` channel which provides pre-compiled binaries for Linux, Windows and MacOS. In fact, it is the recommended channel in the [CVXOPT docs](http://cvxopt.org/install/index.html).

So the final environment file should like:

```YAML
name: Qiskitenv
channels:
 - defaults
 - conda-forge
 - anaconda
channel_priority: true
dependencies:
- python<=3.6
- pip
- notebook
- scipy
- cvxopt
- scikit-learn
- cython
- matplotlib
- sympy
- pillow
- pip:
  - qiskit
  - qiskit-acqua
  - qiskit-acqua-chemistry
```
In this configuration, we specify the list of channels we want to take packages from. The order is important since it acts as a priority list.  
The `channel_priority` option is set to `true` so that packages are selected by priority than by version.  
Please see the [docs](https://conda.io/docs/user-guide/tasks/manage-channels.html) further details.

At the moment, conda doesn't allow  a robust way to specify a channel per package in the environment file so we are left to specify channels globally and hope for the best.  
I will be watching to pull request and test it on my system and let you know if I encounter any problems.

rraymondhp:
Thanks, I really appreciate your help in making this package more robust!  

I will also try to check if the new configuration works. At the moment, we are in the transition phase of changing names, such as `aqua`. I would like to wait awhile waiting for the changes completed. 


```
name: Qiskitenv
channels:
- defaults
- conda-forge
- anaconda
channel_priority: true
dependencies:
- python<=3.6
- pip
- notebook
- scipy
- cvxopt
- scikit-learn
- cython
- matplotlib
- sympy
- pillow
- pip:
  - qiskit
  - qiskit-aqua
  - qiskit-aqua-chemistry
```

ntwalibas:
Instead of speculating on versions and whether things will work as hoped, I went ahead and tested an environment that will work.  
Please find below the YAML file which works without any hitch, as I have tested on my system at least.

```YAML
name: Qiskitenv
channels:
- conda-forge
- anaconda
dependencies:
- python 3.6.*
- pip 18.0
- notebook 5.6.*
- scipy 1.1.*
- cvxopt 1.2.*
- scikit-learn 0.19.*
- cython 0.28.*
- matplotlib 2.2.*
- sympy 1.2
- pillow 5.2*
- pip:
  - qiskit
  - qiskit-acqua
  - qiskit-acqua-chemistry
```
Three things to note:  
- I removed the `channel_priority` for two reasons: one, it is no longer needed for current versions of conda and two, it no longer works anyway and for those who insist on it, they can add it to their `.condarc` file.
- I added versions to every package. As mentioned before, this locks us into a known working version. We can change those versions as things evolve.  
- I removed the `defaults` channel since it is always present by default.

diego-plan9:
Thanks for sharing your impressions @ntwalibas ! Pinning the versions, as in your latest comments, is indeed a way to ensure reproducibility, but with some caveats that I think Rudy will be better suited to evaluate: one is that it introduces another maintenance burden, as the requirements are ultimately defined both in [terra](https://github.com/Qiskit/qiskit-terra/blob/master/setup.py.in#L19) and [aqua](https://github.com/Qiskit/aqua/blob/master/setup.py#L27), and the other is more philosophical: while it does solve it for the "users creating a new conda environment from the .yml and always working in it" case, it does not really tackle the other cases (pip install, users working in a custom environment, etc).

Which actually brings me back to #244 - how many users are experience dependency problems when creating the environment in Python 3.6? We did not have many reports in the terra/aqua repositories or the usual channels - could you elaborate on the error? My line of thinking is that, if it is not really affecting a sizeable number of users, it might be worth adding that to the equation when deciding the way to proceed.

ntwalibas:
As you have noted, pinning versions to each package **independently** of both Terra and Aqua would increase maintain to unsustainable levels in the future.  
Hence, as you rightfully said, Rudy is best positioned to decide how to handle the situation.

As to the philosophical parts, they cease to be philosophical when you end up with a user who is having trouble with pip for instance. Or with custom environments.  
Either way, the idea remains the same: by having versions set and packages listed, if a user is having trouble using pip or say with virtualenv (instead of conda), they can use packages and versions in Conda to see if it helps solve their problem.

Looking at the Aqua and Terra sources you provided (thanks), let us note that both actively specify versions and I believe users will be better served if those versions are used as baseline. The versions I showed above were suggestions after all as I was too lazy to look into the Terra and Aqua setup files.

As to #244, I have attached the relevant errors to the issue, please look at the same.  
The errors will also demonstrate the value of having versions set because conda was detecting that the current environment specifies - implicitly - a different version of `numpy` from that which is specified by both Aqua and Aqua-chemistry.

In conclusion, may I suggest that packages and versions be specified in accordance with Aqua and Terra, for Conda, to get started? Then when you do get reports from users with different setups, you can then see if it is a case of missing package or incompatible version and then make a fix and perform an integration test to see if the latest change works for everyone.  
I suggest the same because if you offer a fix for one user today, you may be offering a fix for ten/hundred users tomorrow, saving yourself future headache and attracting new users. Especially when the fix is extremely simple as is the current case.

As to your line of thinking, I think since quantum computing is new to the masses and there are no financial incentives (yet), some users, instead of filing an issue, will just give up. So can we really rely on the number of filed issues to decide which issues are worth fixing and which are not?

rraymondhp:
@ntwalibas  @diego-plan9  Thank you very much for the active discussion here. I really appreciate your feedback as I cannot perform test on many different machines. As @ntwalibas  pointed out, there are users who just gave up when encountering errors. I indeed witnessed students who had troubled with the installation (MacOSX and Linux). 

IMHO, fixing the version for each package is one of the best options to solve the installation issue. However, as you pointed out that can limit the flexibility of other users. And because the environment also relates to other qiskit packages, I think we should coordinate with other teams too. 

Meanwhile, I would like to take your valuable suggestion by documenting the yaml file in the wiki for `troubleshooting` so that we can refer users experiencing difficulties to it later. Please let me know if you have any other idea.

ntwalibas:
@rraymondhp Apologies for the late reply.  
I am away from my "work station" so I can't contribute to the documentation but in two days I will send a pull request containing the details you mentioned.

Also, I forgot to mention this, but I had tried an installation with virtualenv instead of conda and it failed as well.  
This was due to qiskit terra and aqua and nothing to do with the tutorial. So again, in two days time, I shall file an issue containing the relevant details and link the same here. That is if the issue wasn't fixed in the meantime.

ntwalibas:
Everyone, apologies for being 4 days late to my promise. I was out of town and delayed coming back.

So about the pull request to update the documentation with a troubleshooting guide for installation issues.  
I spent quite some time trying to come up with a reasonable troubleshooting guide and realized that ultimately not only we can't cover all possible situations (and they might keep increasing) but it will create more confusion than offer solutions.  
Therefore, I suggest that users be encouraged to submit issues when there are problems.  
I have access to a Windows 10 machine, an Ubuntu 16.04 LTS and a MacOS Sierra box and can perform tests on those. So far I'm testing on Ubuntu 16.04 LTS and will keep reporting issues.

In that spirit, I had tried an installation using virtualenv (instead of conda) on Ubuntu and it failed for two reasons:
1. psutil: this one failed to install because I didn't have Python development headers installed. The fix is obvious.
2. numpy: this one failed because qiskit specifies numpy>=1.13 while qiskit-acqua and qiskit-acqua-chemistry both specify numpy<1.15,>=1.13.  
I just open an issue for that at [#94](https://github.com/Qiskit/aqua/issues/94).

I think the best that I can do is to create a list of problems I've encountered and their solutions and send them to @rraymondhp so he can decide what's best to do.  
Please let me know how you wish to receive such a document.

PR #236, opened 2018-07-24
Tutorials of comparing strings with quantum state overlaps

rraymondhp:
I wrote a tutorial of comparing a sequence of dna/string with quantum state overlaps. The tutorial shows how to use quantum state initialization and circuit inversions.


qubeat:
Interesting. Could you also provide some reference to the method? Is it something like "amplitude encoding" in recent review by Dunjko & Briegel, Rep. Progr. Phys. 81 (2018) 074001 (e.g. see section 6.3.2)?

rraymondhp:
@attp  Thanks! I have added suggestions to look for other genes as well.

@quantumjim  Thanks, too! I have tried to add explanation at `Deep Dive` section, with links to repetition code and others. 

@qubeat Thanks for pointing out the paper that I was not aware of. I saw the idea first in the Quantum Fingerprinting paper by Buhrman et al. (see the `Deep Dive` section of the updated notebook). A recent paper by MIT-IBM group used this technique for kernel computation.

qubeat:
@rraymondhp Thank you. I myself used similar internal product measure in quant-ph/9703010, 9802028, but the example with encoding into operator using standard qiskit function is good news for me.

PR #242, opened 2018-07-26
Comparing Quantum and Classical Finite Automata

rraymondhp:
This is a notebook created by a student visiting Keio Hub.


rraymondhp:
@jaygambetta  Thanks! I have resolved the conflict.

PR #263, opened 2018-09-16
Quantum Tic Tac Toe

MA0R:
Quantum tic tac toe working in command line. Allows players to make moves in a superposition of several locations in the standard tic tac toe grid. Moves become entangled with past moves, if a player chooses to play in a place possibly occupied by another move.


rraymondhp:
Hi @MA0R 
Thank you very much! Is it possible to run it with jupyter notebook? 
To my understanding the cells in notebook are the same with terminal line. 
It would be great to be able to run your code with jupyter notebook.

MA0R:
Hey @rraymondhp! Yes it can be done, the whole program can be run in one notebook cell, but that's not what you are after I guess. q_tic_tac_toe has a lot of boring stuff, like checking if moves are legal and so on, so I did not think a notebook will be interesting (?). Could it be instructive to write the composite_gate file in a notebook as it involves writing circuits?

quantumjim:
Hi @MA0R. Great game! Thanks for the contribution.

I agree with the idea to run it in a notebook. Even if it is just a single cell with `run q_tic_tac_toe.py`, if you have the output from playing a game in there, then people can get an idea of it even if they don't play

Also, another couple of points:
* I think a user would probably prefer the players to be called 1 and 2 rather than 0 and 1. Also, the X and Y axes are not the way round that I'd expect them to be.
* It would be good to have an explanation of how the game implements the moves using quantum operations. For example, at the end of the game it prints the output from the circuit, followed by the tic-tac-toe grid configuration that corresponds to each bit string in the output. It would be great to have an explanation of how that conversion from bit strings to grids is made.

MA0R:
@quantumjim you are totally I right I got x and y the wrong way around! I changed that, and players are referred to by "player [X/O]" which makes a lot more sense than 0 and 1. I will have a go at writing up a notebook which implements a sample game, it will be a good place to include explanations of the game workings too.

MA0R:
@quantumjim that is all done, thanks for the feedback. I assumed that users know a bit of quantum physics when writing the tutorial notebook.

attp:
Hi @MA0R, this is really cool and I think we should also add it to the index: https://github.com/Qiskit/qiskit-tutorial/blob/master/index.ipynb before merging so that people will be able to find it.

MA0R:
@attp in Index, I made a link to the folder reference/games/quantum_tic_tac_toe, I think it links to the corresponding folder in whichever branch this commit exists

PR #266, opened 2018-09-28
Create code of conduct

rraymondhp:
Following the same standard as qiskit-terra


rraymondhp:
Thanks. It is exactly the same as qiskit-terra as I used the template provided by github.

PR #272, opened 2018-09-29
Update reference/qis notebooks to new registration

nonhermitian:
Notebooks in reference/qis now use the new registration system


nonhermitian:
Merging this myself so that ithers do not duplicate

PR #279, opened 2018-10-05
Updating W State 1 to 3

PierreDC:
The calls to QuantumProgram included a wait argument that had to be deleted in order to have the program running in Qiskit 0.6
Also a style correction in W State 2.


PierreDC:
I am a little lost. My goal is to edit only the W State 1, W State 2 and W State 3 notebooks. I hope this will allow it. It seems to be the case when I open "Files changed" at the top right.

quantumjim:
Looks like that is indeed what is changed. So all good!

PierreDC:
@jaygambetta It seems obvious to me now that it would be better for me to make a more radical change by eliminating QuantumProgram and then submit the pull request with the notebooks tested on qiskit 0.6 to the branch terra_0.6 (temp or final). If you agre, I close this pr.

jaygambetta:
Yeah if you can update that would be great. As the quantumprogram is not supported anymore.

PierreDC:
Closing. It will take some time to update and test the new versions on Qiskit 0.6. Thanks for your support.

PR #300, opened 2018-10-06
binder warning

### LINKED ISSUE #231: Not encouraging users to type credentials into mybinder.org
Hi!

This looks like a nice set of notebooks about a cool topic! I am part of the team that maintains mybinder.org where we see quite a bit of activity for this tutorial.

We try very hard to educate our users that they should never ever type things that they wouldn't tweet to the world into a session on mybinder.org. While individual repositories and notebook creators aren't evil there will come the day when someone creates a notebook that steals credentials or uses some other exploit to do so. For the average user it will be hard to tell if this is an evil or a friendly notebook. That is why we try and get people to never put secrets into sessions.

One of your first notebooks asks for a Bluemix password/key :-/

Is it possible to change to using https://github.com/ProjectQ-Framework/ProjectQ or something like that instead? I think Project Q let's you simulate things on a classical CPU?? Maybe with a comment about how to switch to using bluemix if you are executing things locally.

Is this something that is possible? It would be greatly appreciated. I don't really know much about the whole quantum computing landscape beyond a course I took many years ago as a PhD student :)

jaygambetta:
Fixes #231 by putting in the readme a warning about the binder


quantumjim:
There could also be the route of setting up our own server for Binder or jupyterhub, to allow the easy access option without the security issues. I'll look into it for the future.

nonhermitian:
I also brought that up a while ago and, at least at the time, it was deemed too much work.

PR #301, opened 2018-10-06
Installation

jaygambetta:
Making it clearer how to install the IBMQ provider


jaygambetta:
All will be load_accounts.

jaygambetta:
@nonhermitian can you look now that I merged the master into this. I don't think your concerns apply to this pr.

PR #304, opened 2018-10-07
Index file fix and link fixes

jaygambetta:
Grammar fixes and fix QISKit in links


jaygambetta:
I don’t get your advance comment is this because I accidentally pulled in the change #300

Issue #293, opened 2018-10-06
qiskit - basic


jaygambetta:
## Description

In the basic notebook, we need one showing how to add a 3rd party when JKU comes out I will do this with @diego-plan9 and @yehuda-naveh. 

- [x] Getting started
- [x] The IBMQ provider 
- [x] plot sate visualizations 
- [x] Juypter tools
- [x] 3rd Party Provider

In the visualization less on what a state is just how to plot it and show the interactive ones as well. 

## Your Environment

* QISKit version:
* Python version:

jaygambetta:
Getting started done in pr #305

jaygambetta:
Getting started done in pr #307

PR #307, opened 2018-10-07
Ibmqprovider

jaygambetta:
Finishing getting started IBMQ provider This is part of issue #293


jaygambetta:
should be done but I will run it once more in the morning.

jaygambetta:
The qconfig here is not a qconfig it’s just a file that holds the url as I don’t want to say be them.  I can make it clearer that you just write it but it’s to just there for easy of use.  Read the flow and you see why.

jaygambetta:
Thanks @ajavadia next is visualizations

jaygambetta:
@ajavadia your fix broke.

PR #315, opened 2018-10-09
Add token instructions into getting_started and fix spelling in qiskit notebooks

dongreenberg:
1) Added basic token instructions into getting started -
  a) Added a line pointing new users to the QX API token generate page, just to smooth the flow
  b) Made the instructions compatible with Q Network setup flow, so we can have a single getting started notebook compatible with both QX and Q Network.
Context: For the last x months we've used a separate notebook for Q network setup that we sent to client administrators manually. The qiskit-tutorials getting started notebook wasn't compatible with Q Network before, and many new users encountered errors during setup because they didn't know about our Q Network setup notebook or were simply using the Qiskit tutorials and didn't know the nuances of where to change code for Network usage. This was the source of 3 different escalations from ORNL. The recent registration flow changes make it much simpler to maintain a single "getting started" point for both QX and Q Network users.

2) I copied every notebook in the `qiskit` folder into word to spellcheck them, and fixed spelling errors.


jaygambetta:
Can you sign the cla

PR #319, opened 2018-10-09
community - w-state update

muneerqu:
I updated the W-State notebooks to run on Qiskit 0.6. Following #279 @PierreDC can review them and provide additional changes if needed.


PierreDC:
@muneerqu
 What a beautiful job. I looked at the three notebooks in rendered view and found that you had tested the program on simulator successfully (you even won a car, it seems). A very big thank you for your contribution and I do not see any other changes to make currently. Pierre

qubeat:
May be it is better to use local simulator and comment out account command to make that workable without account information saved on the disk (e.g. in mybinder, etc.)?

muneerqu:
That's a good idea, I can change the backend from IBMQ to Aer. However, in many notebooks `IBMQ.load_accounts()` is uncomment (enabled) so anyone running the tutorials on mybinder (without saving the credentials on disk-encouraged) will get the error:
```python
IBMQAccountError: 'No IBMQ credentials found on disk.'
```
and offline users will get the error:
```python
ConnectionError: Couldn't connect to IBMQuantumExperience server: error during login: ...
```

I'm not sure if we should comment all `IBMQ.load_accounts()` from all notebooks to avoid these errors in such situations!

nonhermitian:
load_accounts replaced the old qconfig register code in the noebooks. This was left uncommented with the thinking that maybe the user would want to run the same circuits on actual hardware. However, this could be improved.

muneerqu:
I agree.

jaygambetta:
can you also edit https://github.com/Qiskit/qiskit-tutorial/blob/master/community/teach_me_qiskit_2018/index.ipynb to move yours to the qiskit 0.6

PR #322, opened 2018-10-10
Initial organisation of community/terra

attp:
- initial proposal for initial organisation of community/terra
- switching order of community/terra and community/algorithms in top index.ipynb and README


attp:
For now, just putting a readme in community/terra with a suggestion of which tutorials to go through to learn qis. This will require a rewrite for some of them. 

In the future, it might be worth writing a couple of tutorials on error correction, to go with @quantumjim's repetition code one. Not sure if that should be in this folder or a community/ignis one.

jaygambetta:
@rraymondhp and @attp can we try to have a policy that we don't approve until the author removes the WIP. Feel free to comment on it and help @attp but this way we can t merge until @attp is done.

PR #327, opened 2018-10-11
Cryptography with BB84 - Update 

muneerqu:
update Cryptography with BB84
added author name in the notebook
added qiskit banner in the notebook
moved author name and link form 0.4.15 to 0.6 list in the index.ipynb


rraymondhp:
Thanks!

PR #330, opened 2018-10-11
W state with tomo in qiskit-tutorial/community/terra updated for 0.6.1

PierreDC:
This version uses Aer and IBMQ and get rid of QuantumProgram.
Will also run with the binder (I hope) for any unregistred user with the qasm_simulator.


PierreDC:
@attp And thanks heaps to you for your quick review and merge. Btw, I tested it right now on the binder and it was OK on Aer qaqs_simulator when skipping IBMQ.load_accounts().

For the arbitrary circuit, drawer(test, style=my_style) always gives an error: DAGCircuitError: 'init is not in the list of basis operations'. I had observed before the update, that in the notebook "using different gates", in the section "Arbitrary initialization", circuit_drawer call has disappeared : I supposed it was for the same reason and that is what justified my personal modification.
Do you know by chance a way to go around this problem?

muneerqu:
Hi @PierreDC, the `error: DAGCircuitError:` is caused by a bug when you use `initialize`. There is an open issue about it here: https://github.com/Qiskit/qiskit-terra/issues/946

attp:
@ajavadia and @jaygambetta - note the difference between the old "quantum_gates_and_linear_algebra" tutorial and the new "using different gates" tutorial that @PierreDC points out above.

PR #345, opened 2018-10-18
Creating a third party provider

yaelbh:
Although it is incomplete, I'll appreciate if you can take a look and see that it's in the right direction at all. @jaygambetta @diego-plan9 

More detailed questions are:
- Should I include a complete description of how Job is used to obtain asynchronous runs? I tried to avoid it, feeling that it was out of scope, but this resulted in quite a dummy job class.
- A similar question about QiskitError - need to inherit from it, as done in Sympy, or can be used directly?

I'm concerned that I find it very difficult and time-consuming to create a new backend as part of the tutorial. I'm afraid that this will be the experience of the users. Examples of sources of difficulty:
- Need to skip the transpiler, otherwise need to satisfy requirements of the transpiler that are often hidden. For example it turns out that the backend must have  at least two basis gates, if not then you do not get a nice error message but a complaint about a code line that splits the basis gates.
- Need to understand the different structures of Qiskit like QObj, Result, BaseBackend, BaseJob: what is mandatory to their constructors, what is their internal structure, which methods must be implemented.
- qobj is both something that can be imported from qiskit (from qiskit import qobj) and a common variable name. There is something called qobj.Result, this is confusing with Result.
- Need to understand the role of classical registers in all this.

Questions about Terra:
- Why is "shots" part of qobj.config and not of circuit?
- Perhaps we could write a class in Terra named AsynchroneousJob, which inherits from BaseJob, and serves as a base class for many jobs. This could help with the code duplication that we discussed.


yaelbh:
It remains to decide how deep to go with jobs, see comment above about it

yaelbh:
Additional sources of difficulty:
- The compilation modifies the circuit, in a way that may not fit the addon.
- The circuit's name (which is not a mandatory field) must be identical to the experiment name for Result to work. This is however not set automatically.

yaelbh:
Also need to decide what to do in this tutorial with classical registers. Maybe remove them altogether.

yehuda-naveh:
@yaelbh is this still in WIP? from your side can we review and merge?

yaelbh:
Not ready to be merged, but I'm not making any progress because I'm waiting for feedback on what's been done so far and answers to questions in previous comments.

yehuda-naveh:
Hi @yaelbh , as we're about to publish the jku/addons blog and we need this finalized as preliminary, let's try to remove the WIP and merge. A first shot at your questions:
- For Job - do the full description, or at least something that would not keep it empty
- Qiskit error - whatever you feel is more friendly for the external addon developer (inherit from base, or direct)
- As for the transpiler/qobj etc questions - I probably don't have the best detailed answers, but in general let's find the best compromise between simplicity to the addon developer, while still having a minimal Terra functionality. We can later improve on this and change the tutorial, but should start with something

A different topic - adjusting to the new Terra. I suggest to start having a tutorial version which is not necessarily up to date to all new Terra interfaces, and then improve/correct it. Presumably the first next addon will be after 0.7, so we'll have stable interfaces by then.

Of course - any corrections / suggestions from anyone are much appreciated, otherwise I'd be happy to have a preliminary version even if to be improved later

yaelbh:
@yehuda-naveh Ready

yaelbh:
Need to update to Terra 0.7. It's on my list as part of the same task for all the add-ons.

yaelbh:
Updated to Terra 0.7, so from my side it is ready for merge.
@yehuda-naveh @jaygambetta

yaelbh:
@jaygambetta Anything else to do here?

yehuda-naveh:
It looks good with me. 

@jaygambetta we're referring to this tutorial from the 3rd-party/JKU blog soon to go out. Can you please merge before we publish?

PR #347, opened 2018-10-19
Add clarifications about using different gates in the terra tutorial

Travis-S:
Based on an email thread with @rraymondhp and @ajavadia, I've added some content to the tutorial on basic quantum operations (file ``terra/using_different_gates.ipynb``) that:

* explains the tensor product operation a bit more
* provides some clarification about how terra thinks about ordering multi-qubit states
* explicitly works through computing the matrix representation of a controlled gate
* adds content on the Toffoli and Fredkin gates

The reason for these changes is that I was confused when working through the tutorial, and figured adding these changes would help prevent others from becoming confused.


jaygambetta:
also can you accept the cla

ajavadia:
Thanks @Travis-S.

The latex under the Mathematical Preliminaries section doesn't render correctly. Can you double check please?

Travis-S:
@ajavadia If you mean "rendering through Github directly looks bad", I agree - see [here](https://github.com/Qiskit/qiskit-tutorial/blob/bef174b83f8e6c5f52a75b4d6312cd8f9236282d/qiskit/terra/using_different_gates.ipynb). However, rendering the NB using the Jupyter NB viewer looks just fine - see [here](https://nbviewer.jupyter.org/github/Qiskit/qiskit-tutorial/blob/bef174b83f8e6c5f52a75b4d6312cd8f9236282d/qiskit/terra/using_different_gates.ipynb).

Also, my local render looks just fine. This appears to be an issue with how Github renders the NB, not the NB itself.

PR #348, opened 2018-10-20
update aqua to 0.3

chunfuchen:
1. update aqua notebooks to aqua 0.3
2. update index for aqua
3. initialize community/aqua

Even though there are 107 changed files, there are about 90 files are just renamed and move to another folder.

Will send another PR for community/aqua


jaygambetta:
Thanks, there is still some organization that i would like to do but this is good.

PR #349, opened 2018-10-21
Fix typo in README.md

ShengliangD:


jaygambetta:
@ShengliangD for the fix, but before we can accept we need you to sign the CLA. If that is possible that would be great.

ShengliangD:
It's interesting to be a contributor in this way  : )

PR #350, opened 2018-10-21
Update README.md

jaygambetta:


jaygambetta:
i dont want it to say general quantum algorithms. It should be Using Aqua if that is.

chunfuchen:
okay, `general` is good, I will update them in my another pr. Thanks.

PR #358, opened 2018-10-24
Pierre dc patch 1

PierreDC:
Fix the problem concerning the graph of the circuit corresponding to the arbitrary  initialization


PierreDC:
I am particularly indebted to Ali Javadi who gave this solution to Anandu in the general discussion of slack Qiskit yesterday

jaygambetta:
@attp can you look at this.

PR #362, opened 2018-11-01
Fourier Transform-Update

muneerqu:
Add: missing figure
Delete: text reference to 'QuantumProgram'
Update: url to 'qiskit/tools/qi/qi.py'
Change: 'ibmqx4' to 'IBMQ device'


attp:
Thanks @muneerqu , I only updated the code and not the text!

PR #371, opened 2018-11-08
[WIP] Add content to relaxation and decoherence tutorial

Travis-S:
This is a PR to add more content and quantum information content to the tutorial on relaxation and decoherence in the ``ignis`` directory. These changes were inspired in part from a discussion on the Qiskit Slack channel on 11/7.

Right now, I've made the following changes:

* Add more description of T1 time and the experiment that measures it.

* Add more description about the T2 star time.

I've tried to base my explanations on this [Quantum Computing StackExchange post](https://quantumcomputing.stackexchange.com/questions/2432/whats-the-difference-between-t2-and-t2/2451#2451). To finish out the remainder of the tutorial, I need answers to the following questions:

* When measuring T2 star, do the oscillations in the outcome probability come from the natural Z rotation induced by natural oscillation of the qubit at its resonance frequency?

* I'd also like to understand the explanation

> In the actual experiment, we change the phase of the pulse before the measurement in order to create oscillations in the observed dynamics. If we just did two Hadamard gates separated by a delay, we would observe a decay of characteristic time 𝑇∗2, but with a strong dependence on any deviation of the calibrated qubit frequency from the actual one. By implementing the qubit pulses with different phases, we shift the frequency dependence into the oscillating feature of the dynamics, and can fit the decaying envelope for a more faithful measure of the coherence time.

* Based on the StackExchange post, I gather that the purpose of the Hanh echo experiment is to eliminate any contributions to the decay of coherences arising from amplitude damping. Is that the right intuition?

I'd be appreciative of comments from @jaygambetta, @nonhermitian, @taalexander as they have more subject-matter expertise on this.


jaygambetta:
@Travis-S the Hahn echo gets rid of fast noise that happens while the qubits are idle. 

The oscillations are fake (Frame change) as it makes it easier to fit a decaying oscillating function vs a decaying function without oscillations. It really is not needed just makes it simpler

jaygambetta:
changes look good.

qiskit/ignis/relaxation_and_decoherence.ipynb
```diff
@@ -422,9 +422,7 @@
     "\n",
     "$$\\Gamma = \\Gamma_{T_{1}} + \\Gamma_{2} + \\Gamma_{\\mathrm{other}}. $$\n",
     "\n",
-    "Phenomenologically, the rate $\\Gamma_{\\mathrm{other}}$ quantifies how other noise processes in addition to amplitude damping and dephasing affect decoherence. Note that this kind of decomposition can always be done if: (a) the noise is Markovian (so that it can be described using _Lindbladians_), and (b) the Lindbladians commute with one another, or they do not commute but the noise strength is small.\n",
-    "\n",
-    "The total decoherence rate sets the rate at which coherences between $|0\\rangle$ and $|1\\rangle$ change:\n",
+    "Phenomenologically, the rate $\\Gamma_{\\mathrm{other}}$ quantifies how other noise processes in addition to amplitude damping and dephasing affect decoherence. A explanation for why $\\Gamma$ can be decomposed in this way is beyond the scope of this tutorial; the reader is encouraged to examine THIS REFERENCE for more details.\n",
```
Travis-S: @taalexander Does this address the concerns your raised on e1fc9c0? If so, is there a good reference to provide?

taalexander: I am a little lost on what you mean by $\\Gamma_{\\mathrm{other}}$. Do you mean another process like amplitude damping (say amplitude boosting to the excited state) which causes an effective decoherence process as well as other dynamics? Or do you mean another pure-dephasing process that results say from slow-stochastic processes or magnetic field inhomogeneities and may be recoverable?

Travis-S: Perhaps I am misunderstanding something. There are many noise processes that could cause decoherence, as you note. T1 and T2, as presented in the tutorial, arise from amplitude damping and dephasing about the Z-axis. Therefore, to incorporate the effects of other noise processes besides those two, I introduced another rate parameter. It's meant to capture decoherence effects from _everything else_.

In the back of my mind, there's a concern that implicit in the way the material is presented right now is the thought "If a noise process is recoverable, we'll do some decoupling pulses and get rid of its effects. Therefore, we don't have to include it in our model of decoherence.". If that's the case, then calling it out in the tutorial would be nice.

taalexander: Ok, I understand now its the latter of the two cases I listed above. I would just make it clear that $\Gamma_{\mathrm{other}}$ is what the Hahn Echo/CPMG sequence are reducing and remove the line 
`Usually, we assume $\Gamma_{\mathrm{other}} = 0$.`

Travis-S: Is the following statement in the right ballpark:

> $\Gamma_{\mathrm{other}}$ captures the effects of noise processes in addition to amplitude damping and pure dephasing about the Z-axis. To reduce this rate, we use  _echo sequences_ that are designed to cancel out the effect of some of those noise processes. As discussed below, echoing out allows us to estimate a more accurate value for T2."

There's a good science question buried in this discussion, that I'd like your thoughts on: suppose I have a noise process that causes dephasing about some goofy axis on the Bloch sphere (i.e, the spindle of qubit states that are fixed points of the noise process is not aligned with the X,Y,Z axes). Two questions:
* Do we say this noise process cause decoherence?
* Does $\Gamma_{\mathrm{other}}$ have a contribution due to this noise process?

If the answers are both "yes", then I'm OK using the quote above. If not, then we'll want to go back to the drawing board.

taalexander: I believe the answer is yes. Provided there is a component of the noise aligned along the Z-axis there will be a component of pure-dephasing that will contribute to $\Gamma_{\mathrm{other}}$ along with other dissipative maps due to the decomposition of the dissipator into its non-commutative Pauli components. Admittedly I am a bit out of my depth here.

jaygambetta:
@Travis-S can you rebase and then put the chances into master

Travis-S:
@jaygambetta Will do. There's still some technical things to discuss. I'll figure out what can go into master directly and what needs to have another PR.

PR #380, opened 2018-11-14
Add tutorial for using the qiksit circuit drawer

mtreinish:
This commit adds a tutorial explaining how to use the circuit drawer and
all the various options for it.

This will depend on qiskit terra 0.7 being released before we can merge it

Still TODO:

- [x] Add section on `qiskit.tools.visualization.circuit_drawer`
- [x] Fix link formatting
- [X] Add link to hosted style option docs (just inlined the list from the docstring we can switch to a link in the future if needed)
- [X] Come up with a less ugly example for using style option with mpl (changed to orange still ugly, but doesn't hurt my eyes as much)


mtreinish:
@jaygambetta ok the current revision has your suggestions. 

As for the warning that's caused by setting `interactive=True` on `mpl` in when running in jupyter. That option isn't meant to be used from within jupyter (although apparently it works on my system at least for `latex), but I put the calls in there still for demonstration purposes. I can remove the code block from the notebook and only refer to it in the markdown about the option.

jaygambetta:
yeah i would put it in the markdown

qiskit/basics/getting_started_with_qiskit_terra.ipynb
```diff
@@ -121,7 +121,7 @@
    "source": [
     "## Visualize Circuit\n",
     "\n",
-    "You can visualize your circuit using Qiskit Terra `circuit_drawer`, which plots circuit in the form found in many textbooks."
+    "You can visualize your circuit using Qiskit Terra `circuit_drawer`, which plots circuit in the form found in many textbooks. For more details on the options when visualizing a circuit see: [Visualizing a Quantum Circuit](visualizing_a_quantum_circuit.ipynb)"
```
jaygambetta: I think this link when we make it html will be broken i would just remove

jaygambetta:
@mtreinish do you need to do the other parts or is this good to take of wip and merge

mtreinish:
@jaygambetta the only thing I think we really need to fix here (besides your recent review comments) is the link to the rendered api documentation for the `style` kwarg. There's a bit of a chicken and egg problem though because we need to release to get that link published (since it's one of my recent visualization doc fixes) and we also want this notebook in the published docs for the release. Right now it just says **INSERT LINK TO RENDERED DOCS HERE** where I was planning to put that.

I also need to add a bit of text about installing matplotlib for the mpl output now that it's optional.

But after all that I think this should be good to go.

qiskit/basics/getting_started_with_qiskit_terra.ipynb
mtreinish: Done

mtreinish:
@jaygambetta ok, I've updated this per the review comments and just in-lined the list of options for the `style` dict with mpl instead of linking to them. So I think this is good to go now. The only thing is we'll have to rerun the notebook to regenerate the images after we fix the bit ordering with https://github.com/Qiskit/qiskit-terra/pull/1280 https://github.com/Qiskit/qiskit-terra/pull/1278 and https://github.com/Qiskit/qiskit-terra/pull/1276

PR #388, opened 2018-11-18
Plotting data in qiskit

jaygambetta:
@mtreinish please edit this but I made this replace the other plotting notebook. 

@attp could you have a look.


jaygambetta:
I think so. If you can fill it out how you think that would be great.

mtreinish:
I did add a docstring for all those options in https://github.com/Qiskit/qiskit-terra/pull/1229 so it should make adding an explanation for the various options a bit simpler. I can take a look at adding that here

jaygambetta:
@mtreinish great. Please add.

mtreinish:
I've got the updates pending here: https://github.com/jaygambetta/qiskit-tutorial/pull/1

jaygambetta:
Thanks @mtreinish. Do you think we should merge or leave them as pr until 0.7 is out

PR #404, opened 2018-12-04
Switch quantum emoticon to use 7-bit ASCII and ibmq_melbourne_16 (fixes #41)

holly-cummins:
* Fix some deprecation warnings related to using global functions rather than the new `IBMQ` ones
* Use 7-bit ASCII and a 14-bit register, in order to enable ...
* ... using `ibmq_melbourne_16` as the backend rather than the retired `ibmqx5`


holly-cummins:
Good spot! Those are now fixed, thanks.

PR #409, opened 2018-12-08
update the grover+evolution+eoh+vqe2iqpe+h2qpe notebooks

hushaohan:
also, remove the grover.py file


community/aqua/chemistry/h2_qpe.ipynb
```diff
@@ -6,23 +6,144 @@
...
+    "from qiskit.transpiler import PassManager\n",
+    "from qiskit_aqua import AquaError\n",
+    "from qiskit_aqua import QuantumInstance\n",
+    "from qiskit_aqua.algorithms.classical import ExactEigensolver\n",
+    "from qiskit_aqua.algorithms.components.iqfts import Standard\n",
+    "from qiskit_aqua.algorithms.single_sample import QPE\n",
+    "from qiskit_aqua_chemistry.drivers import ConfigurationManager\n",
+    "from qiskit_aqua_chemistry import FermionicOperator\n",
+    "from qiskit_aqua_chemistry.aqua_extensions.components.initial_states import HartreeFock\n",
+    "from qiskit_aqua_chemistry.aqua_extensions.components.initial_states import HartreeFock\n",
```
chunfuchen: redundant import

chunfuchen:
for declarative approach (using JSON input), please also use backend object rather than using backend as a string. 
to do that, just remove the backend name in the JSON but keep other setting, and then pass the backend object get from Aer or IBMQ to `run_algorithm` or `.run` method.
like:
```python
run_algorithm(params, algo, backend=backend)
```

hushaohan:
> for declarative approach (using JSON input), please also use backend object rather than using backend as a string.
> to do that, just remove the backend name in the JSON but keep other setting, and then pass the backend object get from Aer or IBMQ to `run_algorithm` or `.run` method.
> like:
> 
> ```python
> run_algorithm(params, algo, backend=backend)
> ```

how to pass in `shots`?

chunfuchen:
@hushaohan you can still specify the setting related to backend in the `backend` field of the `params`. only the `name` field is removed. 

like
```python

params = {'backend': {'shots': 100},  others...}
backend = Aer.get_backend('qasm_simualtor')
run_algorithm(params, algo, backend=backend)

```

chunfuchen:
@hushaohan Aqua is updated for the import hierarchy.
you can directly do `from qiskit_aqua.algorithms import XXX`

PR #418, opened 2018-12-13
Update circuit drawing notebook for order changes again

mtreinish:
At the last minute we decided to change the order back to what it was
before, and standardize on that for visualization. This commit adjusts
the visualizations in the notebook to that, by running with
 Qiskit/qiskit-terra#1475 applied.


mtreinish:
This depends on https://github.com/Qiskit/qiskit-terra/pull/1475 being merged, so we should wait for that before moving forward with this.

jaygambetta:
approving as it has a warning but for more changes lets put to terra branch. I am going to rebase that and then remove the warning.

PR #419, opened 2018-12-14
Add readme for notebooks in terra/qis_adv

rraymondhp:
Adding one-liner or two explaining the notebooks under qis_adv


jaygambetta:
@attp im good with this but im not part of the approving for community

PR #426, opened 2018-12-15
update imports for several `community/aqua` notebooks

hushaohan:
for grover+evolution+eoh+vqe2iqpe+h2qpe+h2iqpe notebooks under `community/aqua/*`

### DIFF
--- a/community/aqua/chemistry/h2_iqpe.ipynb
+++ b/community/aqua/chemistry/h2_iqpe.ipynb
@@ -21,7 +21,7 @@
    "source": [
     "import numpy as np\n",
     "import pylab\n",
-    "from qiskit import Aer\n",
+    "from qiskit import LegacySimulators\n",
     "from qiskit_aqua_chemistry import AquaChemistry\n",
     "import time\n",
     "\n",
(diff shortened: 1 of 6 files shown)


community/aqua/chemistry/h2_iqpe.ipynb
```diff
@@ -48,7 +48,7 @@
     "]\n",
     "\n",
     "backends = [\n",
-    "    Aer.get_backend('qasm_simulator'),\n",
+    "    get_aer_backend('qasm_simulator'),\n",
```
jaygambetta: why are we wrapping up backends

pistoia: @jaygambetta , we created that wrapper because Aer is not on GitHub yet, and we cannot use it directly and be sure that the code we write now will work, so the wrapper implements a fallback.  We need that wrapper to test Aqua with Aer whenever Aer is installed, which is good because it crashed in one unit test.  However, we did not mean for that wrapper to be used in public tutorials.  We are going to use Legacy for now.

jaygambetta:
@attp could you review these

PR #438, opened 2018-12-18
move monitoring notebook to terra

nonhermitian:
Also add notebook to main index.ipynb


nonhermitian:
This is a mess will just redo

PR #460, opened 2018-12-21
Adding README and exercises to qiskit/terra

rraymondhp:
Adding README to explain what each notebook is about. 
Adding exercises to introduce some basic concepts of quantum circuits.

### DIFF
--- a/qiskit/terra/README.md
+++ b/qiskit/terra/README.md
@@ -1,0 +1,24 @@
+# Operating Quantum Circuits with Qiskit Terra
+
+This folder hosts notebooks on the basics of Quantum Circuits with Qiskit Terra.
+Running quantum computation with Qiskit Terra starts with building quantum circuits, which
+consist of quantum registers, classical registers, and quantum operations.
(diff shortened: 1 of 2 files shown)


jaygambetta:
i think these should go in the community folders.

rraymondhp:
Jay, I was thinking of writing exercises corresponding to notebooks at each subdirectory. I thought it would be easier for users to look for hints if each exercise placed under the same subdirectory. What do you suggest? If we place exercises under community, I wonder if they should be under a single subdirectory dedicated to exercises.

jaygambetta:
I think this is a good idea to have an exercise. For the qiskit directory, i want it to autogenerate into the documentation.

rraymondhp:
To clarify, README.md will be autogenerated from `index.ipynb`? 
I think we need a guide (README.md or index.ipynb) about how to traverse in sequence notebooks at each subdirectory. 

It is not trivial where to start when we visit the subdirectory qiskit/terra, for example: 
https://github.com/Qiskit/qiskit-tutorials/tree/master/qiskit/terra

PR #463, opened 2018-12-21
Split monitoring tools into Terra and Jupyter sections

nonhermitian:


jaygambetta:
in the Jupyter one 

Monitor the job using job_monitor in blocking-mode (i.e. using the same thread as the Python interpretor)
 
should this be in non-blocking mode

nonhermitian:
This should be good now.

PR #474, opened 2018-12-27
Quantum slot machine 

nonhermitian:
A simple quantum slot machine.


nonhermitian:
@quantumjim It has been moved

PR #482, opened 2018-12-28
Update slot machine to payout and wait on IBMQ backend job

nonhermitian:
### Summary
Addresses nit picking from @jaygambetta.


nonhermitian:
@quantumjim it seems I need your approval.

PR #503, opened 2019-01-22
Fix 'Bell state' emoticon tutorial url

juanfran:
Changing the 'Bell state' url in the emoticon tutorial. The old path doesn't exist.


jaygambetta:
@quantumjim can you approve.

PR #508, opened 2019-01-31
Add noise model dict

quantumjim:
A command is added to the end of the qiskit/aer/device_noise_simulation.ipynb notebook. This prints out the noise model as a dictionary, to show a pretty comprehensive example of what a noise model looks like.


quantumjim:
This more than doubles the length of the notebook, because the dict is pretty big, but it's all dumped at the end.

Issue #523, opened 2019-02-19
move creating_third_party_provider.ipynb from basic to terra folder


jaygambetta:
### What is the expected behavior?
This we put in the wrong place. 

Also, we need to update the index file to link to it.

PR #524, opened 2019-02-19
Moved creating_third_party_provider.ipynb from basic to terra folder

yaelbh:
### Summary

According to #523


yaelbh:
Done.  Do we prefer "third party" or "third-party"?

Issue #528, opened 2019-02-21
An output of entanglement introduction.ipynb in qiskit terra tutorial might be wrong


ChenChengKuan:
### Informations
I use binder image in the Qiskit readme
- **Qiskit version**:
- **Python version**:
- **Operating system**:

### What is the current behavior?
The histogram of cell 13* in entanglement_introduction.ipynb shows only one state (00), however, it should show 2 random state as described in the context. 

*The output is the measurement of the figure below
<img width="400" alt="2019-02-21 7 21 19" src="https://user-images.githubusercontent.com/13601253/53173256-466f5480-3622-11e9-9e8d-e9fadbcd16e0.png">

### Steps to reproduce the problem
I get the same result every time when I reset and restart the notebook

### What is the expected behavior?
I should show random states as described in the context instead of a deterministic state.

fvarchon:
Hello,
I use (qiskit 0.7.2| python 3.6.3 |Anaconda custom (64-bit)| Windows 7). The problem seems to be related to BasicAer.

I am able to reproduce the same bug when I use:
backend = BasicAer.get_backend('qasm_simulator')

The problem disappears if I use: 
backend = Aer.get_backend('qasm_simulator')

ChenChengKuan:
Hello
it works like magic! , Thanks for your answer

PR #530, opened 2019-02-22
fix small bugs described in #528

ChenChengKuan:
### Summary
I replace BasicAer.get_backend('qasm_simulator') with Are.get_backend('qasm_simulator') to fix the bug described in #528


jaygambetta:
closing as basicaer will be fix in 0.7 update.

PR #535, opened 2019-02-27
Adding community/exercises folder and an example of exercise notebook

rraymondhp:
### Summary
I am proposing to add a directory that contains exercises to help users walk through tutorials and see if they understand what are written on the notebooks. 

### Details and comments
I added `exercises` directory under the `community` directory that will host notebooks containing exercises. The first exercise is also added. Comments and suggestions are welcomed!


rraymondhp:
@quantumjim  Thanks!
and ups, my bad! It was my mistake to include the answers.

But you are right, we must find a way to provide (and hide) the answer and cope with questions that have many possible answers.

Issue #554, opened 2019-03-09
error in week 0, slide 9,


omishina:
Hi all,
Great resource!
I am a quantum physicist and a physics education researcher and educator. 
I am read through week 0 of “Teach me quantum” and I spotted a typo on slide 9: in place of “macro” it is most likely intended to be “micro”.
Best!
Oxana

qubeat:
Hello, I just look at [the slides](https://github.com/Qiskit/qiskit-tutorials/blob/master/community/awards/teach_me_quantum_2018/TeachMeQ/Week_0-Hello_Quantum_World/slides.pdf)? It seems to me, that even after change "macro" to "micro" it still may sound not very clear.

omishina:
Hello,
Thanks for sharing.
I mentioned “Macro” on slide 9 as it is really a typo to change. 
In fact, I also think that there is an alternative, more transparent way for making quantum vs classical distinction.
As for using micro vs macro to distinguish quantum vs classical, this is quite common. It is true that the quantum effects, i.e. observations  that can not be described using classical theory, up to date are predominantly observed in micro systems. Although, it does not mean that quantum theory does not apply to macro objects. The thing is that observation predicted by quantum theory some times coincide with the prediction of the classical theory due to the decoherence in macro systems.
 But! 
With scientific progress we discover more and more systems which are quite macro that innfact demonstrate quantum properties. For example, a quantum memory for light which is build of a gas cell of few cm long (quite a macro object) stored the quantum state and permits to retrieve the quantum light at a later time. In this case, the “micro” vs “macro” division with a given border size does not apply any longer. 
I would say that such a division is rather qualitative, a historical trend, which reflects that in outcdaly live we do not observe quantum phenomena with our own eyes. But even this is questioned by scientific investigation today. One can not really putting a size border as Thierry will always be possible to find an example of a bigger system they demonstrates quantum properties.
To my point of view a more clear classical vs quantum destinction is the one which distinguish the models, but not the systems to which they apply.

yehuda-naveh:
Indeed, superconductivity e.g. a prime example of pure quantum at any macro scale. I made the small fix at #557 . Thanks for noticing the typo

omishina:
Hi,
I looked at the slides again but I see no changes.

yehuda-naveh:
It will be visible after the latex is compiled

jaygambetta:
[...]

PR #557, opened 2019-03-09
micro/macro typo and clarification in Hello Quantum World

yehuda-naveh:
fixed typo and addressed the legit discussion of #554


omishina:
Hi,
I looked at the slides again but I see no changes.

PR #560, opened 2019-03-10
typos on tutorial and binder issues

omarcostahamido:
### Summary
qiskit/basics/getting_started_with_qiskit.ipynb
a couple minor typo corrections

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ In this repository, we've put together a collection of Jupyter notebooks aimed a
 
 The notebooks for these tutorials can be viewed here on GitHub. But for the full experience, you'll want to interact with them!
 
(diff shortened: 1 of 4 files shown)


omarcostahamido:
I have a new pull request to make, I think I fixed the binder issues.
what I had on my clipboard:

> i think i could fix the binder issues. simply by using lowercase on the username "qiskit" and correcting the the path to (...)tutorials instead of (...)tutorial forced it to rebuild the binder image.
> :)
> if this alone is not enough then we may try to specify python==3.6, because one guy here (https://gitter.im/jupyterhub/binder) was able to make it work just by doing so.
> 

and yes, I think python==3.6 is really necessary now.

omarcostahamido:
thank you @jaygambetta 
OCH

PR #566, opened 2019-03-19
Bug Fixing on jupyter notebook example

efviodo:
### Summary
Fixing cell description on backend filtering example at the_ibmq_provider.ipynb notebook.

### Details and comments

I changed a cell description on the filtering the backends example because it had an inconsistences between the cell description and the source code on cell.

I changed the qubits limit description from "have more than 10 qubits" to "have at most 5 qubits" in order to make sense with the condition in source code at the belowing cell. 

I changed cell description and not source code because below in notebooks, there are examples filtering backends with exactly 5 qubits. So in this way the fix makes the notebook more armonic.

### DIFF
--- a/qiskit/basics/the_ibmq_provider.ipynb
+++ b/qiskit/basics/the_ibmq_provider.ipynb
@@ -543,7 +543,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


efviodo:
I'm going to close the PR and create a new one, since I'm made a commit with a different author's name than my github account.

Issue #407, opened 2018-12-07
Bitstring encoding new tutorial


lukasszz:
Basing on the tutotrial "qiskit-tutorial/community/hello_world/string_comparison" I wanted to prepare something simpler and explained in more details. Could you help me answer my questions and provide better explanation in placess marked by: (???)
in [bitstring_encoding.ipynb](https://github.com/lukasszz/qiskit-tutorial/blob/bitstring_enc/community/hello_world/bitstring_encoding.ipynb)

quantumjim:
**??? 1**

In this encoding a four bit output of `abcd` means that the value at position `abc` of the original bit string is `d`. So, for example, `0001` means that the value at position `000` (i.e. 0, and so the first bit in the original string) is `1`. 

To decode, you can loop through all the four bit strings, extract the positions (from the first three bits) and the values (from the fourth) and use them to reconstruct the original 8 bit string. For example (note: untested code)

    decoded_string = [ '' for _ in range(8) ] # list with blank spaces for 8 bit values
    for string in st: # loop over all 4 bit strings in result
        pos = int(string[0:3]) # get integer that corresponds to first three bits
        decoded_string[pos] = string[4] # set the value of the decoded string at this position to the fouth bit
    decoded_string = ''.join(decoded_string) # turn list of 8 bit values into 8 bit string

Ideally, all four bit strings should share the same probability. And it should be impossible to get the results `abc0` and `abc1`, because this would be telling us that the bit value at position `abc` is both `0` and `1`: not possible for a classical string.

In real life, imperfections happen. So you could use the probability to resolve ambiguities: if you find both `abc0` and `abc1` in your results, accept the most probable and reject the other as noise.

**??? 2**

The initial state of any circuit is that with all qubits in state `0`.  So `encode_bitstring(BS1,...)` takes `0000` to the required superposition  for BS1, and `encode_bitstring(BS1,...,inverse=True)` would take that superposition  (and only that one) back to `0000`. This is why the combination of the two yields `0000`.

**??? 3**

Let's use |BS1⟩ to denote the superposition state created by `encode_bitstring(BS1,...)`, and |BS2⟩ for the same with BS2. For these states, as with any states, we can write

|BS2⟩ = *a* × |BS1⟩ + *b* × |other_stuff⟩,

[...]

PR #575, opened 2019-03-24
New tutorial Compressing bit strings in qubits using superposition 

lukasszz:
### Summary
Basing on the tutorial "qiskit-tutorial/community/hello_world/string comparison" I wanted to prepare something simpler and explained in more details. So I created new tutorial which is designed to be an independent teaching material.

#407


lukasszz:
Thank you @quantumjim.
I've made all your suggestions and do the additional typos and grammar check.

I also change the `initialize()` call on the 13th cell, as I noticed it can just take a `QuantumRegister`.  I think we should also upgrade @rraymondhp tutorial, and probably do better documentation or parameter naming in Terra.

PR #577, opened 2019-03-28
Add quantum animations

quantumjim:
### Summary

Notebook using similar methods to the terrain generation notebook. This one has a method of loading images into states, manipulating them and producing animations.

### Details and comments

Uses the `PIL` and `apng` packages, but I don't know if we want to add them into the environment necessarily.


quantumjim:
@attp  Thanks for the feedback. I'll act on it soon.

quantumjim:
@attp done!

PR #587, opened 2019-04-11
Qiskit Arts subdirectory of Qiskit Community tutorials

alexiskirke:
### Summary

I would like a subdirectory on application of Qiskit to the arts in the community section. I have added a first tutorial about superposing music in Windows using Qiskit. 

### Details and comments

I foresee a growth in applications of Qiskit to the arts and am working on a number of projects myself. Also the creation of art using Qiskit is a great way to learn about basic quantum computer functions. One issue I have found is that the Notebook I uploaded won't render in Github, but it will render in other online renders. Github does have a repeated issue with rendering from 2016 to now: e.g  https://github.com/jupyter/notebook/issues/3035

I have plans for further notebooks, and also suspect that others will add. I will also create notebooks for my contributions to the Qiskit Camp Hackathon.

Alexis Kirke


aasfaw:
There are several demos that could fit in here. +1

alexiskirke:
Hmmm looks like I placed this in too high a directory. I’ll move it into community

attp:
@quantumjim what do you think about having an _art_ directory in _community_? could be a better place for your pixel art tutorial and terrain generation. or do you think it would make more sense to merge the _games_ and _art_ directories together?

alexiskirke:
I did wonder about that myself @attp , but I think there's a possible issue that it will start to create a dichotomy between quantum applications and "non-serious" quantum applications - and that all non-serious applications will end up under "games". This will become confusing eventually, and perhaps also be a de-motivator for those doing "serious" research into non-scientific/engineering applications of QC?

quantumjim:
I have also wondered about this. Specifically, I thought of renaming 'games' to 'creative' and broadening the scope.

It is probably a good idea to keep games and art folders separate, though. One of the purposes of the games folder is to be a bit non-serious. As @alexiskirke says, we don't want people to get that impression for all the creative applications.

So I agree with having an art folder separate from games, and I would also move my terrain generation and pixel art into it. 

In the long term, though, perhaps we should also start thinking of whether these kinds of projects could form the basis of a fifth element: Qiskit Mousa, perhaps?

alexiskirke:
That could be a great idea. What's Mousa? I'm guessing you're not referring to the Scottish island....
I've recreated an "arts" directory but within Community. Should I pull this pull and create another pull?

rraymondhp:
I believe the fifth element is called "aether"...

quantumjim:
My 'Mousa' was an attempt to use the Greek for 'muse', in keeping with the classical theme. But it doesn't have the same recognition as the latin words for the four elements.

jaygambetta:
after #616 lands @aasfaw, @rraymondhp, @attp  and @quantumjim should be able to approve this and merge. If you have issues please let me know and I can see what the issues is

quantumjim:
I've been planning to do a few edits to this PR (moving some notebooks from the games folder, changing READMEs, etc). But now `CODEOWNERS` has been changed, maybe it's better to that in a separate PR.

The only thing I'd suggest changing for this PR is to expand the README, and make it more in line with something like [this one](https://github.com/Qiskit/qiskit-tutorials/blob/master/community/games/README.md).

omarcostahamido:
very excited about this :)

PR #607, opened 2019-05-02
Pulse calibration notebook

taalexander:


jaygambetta:
can we not put this is in pulse folder and can we link it in index. 

Also you may want to delete the errors with your user name

PR #627, opened 2019-05-07
Move Hello_Quantum to qiskit/basics

quantumjim:
As discussed with @jaygambetta on Slack, Hello_Qiskit.ipynb and associated helper functions have been move to qiskit/basics.


quantumjim:
It's being moved because it needs to be in the qiskit folder to be accessed by IQX.

I'm not sure about adding a number, since it is not part of the series formed by the other notebooks. Having it unnumbered, and sorted below them, seems more natural to me.

quantumjim:
I dismissed @rraymondhp's review in the hope of getting this in today so I could test if it worked. Seems that approval for @ajavadia or @jaygambetta is still needed, though.

I should note that @rraymondhp's suggested edits should definitely be taken into account. Maybe not for the initial commit, but certainly soon.

rraymondhp:
Hi @quantumjim 
Thanks for the consideration. I recalled that the purpose of basics is to group simple notebooks. I was not aware that the notebook you moved is needed for IBMQX.

quantumjim:
It seems like it won't work in the new Quantum Experience currently, so shelving this PR.

PR #631, opened 2019-05-07
Codeowner update for changed aqua folders

woodsp-ibm:
Update to CODEOWNERS file for changes to Aqua folders where application folders, such as finance, chemistry etc, have been moved out from being sub-folders of aqua. Separate PRs will follow shortly for tutorial notebook updates.


jaygambetta:
@manoelmarques can you remove the qiskit (as there is a more discussion I need to have as we are seeing issues with the new IQX). First I want to get the community so please split this in two.

PR #650, opened 2019-05-11
Random Number Generator in Range (a,b)

QuantumNovice:
### Summary
Algorithm to generate random numbers in a specified interval  (a, b) using N qubits, where N >= 2.


rraymondhp:
Hi,
Thanks for the contribution. Because you created a new cryptography subdir, it raises a good question if the random number generated by your method is really random. Notice that the devices have noises and I think it is not trivial to show the randomness resulting from an algorithm that tries that. 

There have already papers showing how to generate random numbers from quantum devices by elaborating Bell inequalities and other primitives of quantum information. You may want to look into those papers.

QuantumNovice:
The random number is directly generated from Hadamard gate. It is similar to rolling an n-faced dice, where n is the number of Qbits. It isn't the worst way to generate  random numbers using a quantum computer but the sequence developed is truly random.

rraymondhp:
Hi @QuantumNovice 
For perfect quantum devices, Hadamard gates may generate truly random numbers, but because we have not such devices right now, I doubt the numbers generated are uniformly at random. I wonder if you can try compare the comparison of integers you have. 
If you use simulators you will get pseudo-random numbers that you can determine once you know the seed.

rraymondhp:
Hi @QuantumNovice  and @aasfaw 
I found out that there is already a tutorial on how to generate random numbers here:
https://github.com/Qiskit/qiskit-tutorials/blob/master/qiskit/aqua/generating_random_variates.ipynb

aasfaw:
Hi @QuantumNovice and @rraymondhp. 

As @rraymondhp points out, there is already a `qiskit/aqua` tutorial showing how to use `aqua` to generate random numbers. This is a reasonably good contribution from the community, and note that it is going into the `community/` folder. It will hopefully inspire more crypto contributions.

Regarding the noise from devices -- note that the contribution is using an `Aer` simulation without device noise modeling. It's challenging to argue that the distribution is truly random with real device noise, but this notebook isn't doing that. 

@QuantumNovice can you update with a note pointing to the `qiskit/aqua` method mentioned above? It generalizes your results for several other distributions that one might seek.

aasfaw:
I added the existing aqua tutorial into the readme for this community submission. It's ready to merge.

PR #651, opened 2019-05-13
fixing links: binder and second notebook

omarcostahamido:


omarcostahamido:
the binder was going to a misleading 404 :/

PR #674, opened 2019-05-18
add optimization_level=0 for T1/T2 notebook

ajavadia:
### Summary
We don't want any optimizations in the T1/T2 circuits. Optimization can remove identities, changing the expected behavior (https://github.com/Qiskit/qiskit-ignis/issues/197)


ajavadia:
I fixed the T1/T2 notebook here: https://github.com/Qiskit/qiskit-ignis/issues/197
Are there other parts of ignis that behave like this?

PR #712, opened 2019-06-11
add repetition code tutorial using ignis tools

quantumjim:
### Summary

Tutorial demonsrating the `RepetitionCode` and `GraphDecoder` features of Ignis.

PR #713, opened 2019-06-11
Remove qec section from community/terra now that we have ignis.

attp:
Particularly given PR #712.


attp:
@quantumjim this is mainly removing your previous repetition code tutorial.

PR #718, opened 2019-06-14
add repetition code notebook

quantumjim:
### Summary

Add notebook for how to use the ignis repetion code with real devices.

### Details and comments

Replaces the outdated content that was removed in #713


quantumjim:
The repetition code notebook is the main part of this PR, though I did get a little carried away with fixing a few small typos here and then when adding it to the index. Sorry for all the notification emails that may have generated!

If I understand correctly, this requires @ajavadia or @jaygambetta approval since it involves changes in several folders (though they are all within 'community').

PR #720, opened 2019-06-17
Add process fidelity

quantumjim:
### Summary
As well as calculating fidelity using `state_fidelity`, the calculation is also done with `process_fidelity`.


quantumjim:
Done to answer [this question](https://quantumcomputing.stackexchange.com/questions/6326/how-to-calculate-the-fidelity-of-a-certain-gate-of-a-ibmq-device-in-qiskit-using), after discussion with @dcmckayibm and @chriseclectic.

dcmckayibm:
If you run it, it gives the same answer as before?

quantumjim:
The state fidelities are still in there and give answers that are close to those before. I'd say they are the same, within the bounds of the statistics.

The process fidelities are similar but distinct.

dcmckayibm:
The 2Q examples, the state needs be divided by 4

quantumjim:
Good catch. Fixed!

PR #743, opened 2019-07-23
Purity RB tutorial

ShellyGarion:
### Summary
Purity RB tutorial 

### Details and comments
Copied the notebook from: https://github.com/Qiskit/qiskit-ignis/tree/master/example


ShellyGarion:
@dcmckayibm - please review

dcmckayibm:
I would simplify the `rb_pattern` to just a single pair of qubits. That way the notebook will run faster. When adding the noise it says noise on cx and u gates but there is only noise on the cx gate (I would give it that way, but update the text)

dcmckayibm:
Maybe, also done print out all the circuit names and draw all nine circuits. Select one circuit to draw and print out the circuit names for one length

ShellyGarion:
Add measurement mitigation similarly to the tomography tutorial:
https://github.com/Qiskit/qiskit-tutorials/blob/master/qiskit/ignis/state_tomography.ipynb

ShellyGarion:
@dcmckayibm - I updated the tutorial following your comments

dcmckayibm:
Can you update the beginning (last updated and versions)

dcmckayibm:
"To illustrate, we print the circuit names for purity RB (for lenght=0 and seed=0)" -> fix typo

dcmckayibm:
In the measurement noise section add some text describing why we might need measurement mitigation with purity RB. Also in that section you don't need to also include the coherent error example.

dcmckayibm:
In the coherent error description add something like:

"We define a coherent noise model for the simulator. In this example we expect the purity RB to measure no errors, but standard RB will still measure a non-zero error. "

ShellyGarion:
@dcmckayibm - Thanks for the review - I updated the tutorial following your comments

dcmckayibm:
I think this code 

```
counter = 0
for rb_seed in range(len(rb_purity_circs)):    
    for d in range(npurity):
        correct_purity_result_list.append(meas_fitter.filter.apply(meas_purity_result_list[counter], 
                                                                   method='least_squares'))
        counter += 1
```
just needs to be
```
for meas_result in meas_purity_result_list:    
    correct_purity_result_list.append(meas_fitter.filter.apply(meas_result))
```

The least square is default so you don't need to include that

PR #747, opened 2019-08-04
Ignis logging tutorial 

eliarbel:


nonhermitian:
Can you move this to the community tutorials ignis folder?  It is a but too specialized for the general tutorials.

nonhermitian:
Merge away

dcmckayibm:
Can you format the header to be like the other tutorials?

eliarbel:
Qiskit image added to the header - let me know if anything else is needed.

nonhermitian:
Can this show an example of how this is actually used, i.e. when is this logging functionality beneficial to the user?

dcmckayibm:
> Can this show an example of how this is actually used, i.e. when is this logging functionality beneficial to the user?

It's kind of alluded to in the example notebook because there's a preloaded log file with T1 data. The idea being that everytime you run T1 it will log that number (and for the other characterization methods, e.g., RB). The next step plan is to put those logging commands into the ignis fitters so it will log automatically (if the option is set to have a log file, otherwise it will ignore those commands). We haven't done that yet, but that's the next step.

nonhermitian:
Is this functionality going to be ready for qiskit 0.12?  Otherwise this would be a tutorial teaching a feature that is not implemented yet.

dcmckayibm:
The logging is implemented, so anyone can activate their log file then put logging commands into their own notebooks. Automatic logging is not yet activated, but that's not mentioned in this tutorial.

PR #749, opened 2019-08-05
Updated randomized benchmarking notebook

yaelbh:
### Summary

The notebook has changed:
1. The unitary simulator returns the same result, but with phase.
2. Fitter error for EPC in Ignis had a bug and was fixed - the notebook needs to run with the fixed code (previous EPC fitter error was too optimistic).

### Details and comments

@dcmckayibm numbers of u1 gates per Clifford etc. are very different from the previous version - is it OK ?


dcmckayibm:
Can you update the "last updated" 

Number of U1 gates, I think I'm ok, it's probably reflective of increased optimization in the transpiler.

I think the unitary simulator always returned with a global phase (see my comment in that section).

PR #771, opened 2019-08-19
Numerate and cleanup Aer tutorials

nonhermitian:


jaygambetta:
we still need to add a Clifford stabilizer one. can you make an issue for that

PR #772, opened 2019-08-19
Make tutorial tests run all notebooks by default and configurable

mtreinish:
### Summary

This commit updates the often forgotten about test module to actually be
runnable by default. It also makes the hard coded values for path,
kernel, and timeout configurable via env variables so people can adjust
these easily without having to hand edit code. The exclusion list option
was not added because doing that in a test module is redundant a decent
test runner will support test selection in a more robust manner and that
should be leveraged instead of exclusion list.

In addition because the notebook execution can be slow this commit also
adds a .stestr.conf file so that we can run the notebooks in parallel
(just like we do for the terra unit tests). This can be done with
'stestr run' and by default it will run as many processes as you have
cpus to run a notebook each (assuming you have stestr pip installed
along with the notebook dependencies).

### Details and comments


mtreinish:
This will be useful to automate running notebooks as a post-merge CI job. I'm working on a prototype for doing this here: https://github.com/Qiskit/qiskit-terra/pull/3007

jaygambetta:
how long will this take with the experiment runs. I think we need some time before we add this.

mtreinish:
The experiment runs are definitely slow, probably too slow for CI. In Qiskit/qiskit-terra#3007 it fails on anything that touches ibmq right now. I also had to skip all the aqua notebooks because each one was also super slow, taking >10mins each for many of them. (also one aer notebook was skipped because of CI memory limitations)

But, this doesn't actually add CI or anything. It just updates the test file so that it can be used locally without any modification. Any CI usage will come after when we can figure out how and what we can do sanely in the CI environment.

nonhermitian:
This is good.  Someone could run this and leave it over night or something.  The current manual execution is becoming problematic.

nonhermitian:
@mtreinish can you fix the conflict here then I can merge (no rush)

PR #782, opened 2019-08-21
fix advanced circuits notebook after terra bind params fix

ajavadia:


nonhermitian:
This has a conflict

microftech65:
.pstch

PR #783, opened 2019-08-21
Add transpiler logging section to using the transpiler

mtreinish:
### Summary

As part of the upcoming 0.9 terra release the logging from the
transpiler has been improved and standardized around python standard lib
logging. This tutorial adds a guide on how to leverage this in a couple
of common use cases to document how someone can configure the logging
from the transpiler.

### Details and comments


mtreinish:
I wasn't sure which transpiler notebook to put this in, I can easily move it if this isn't the right fit

mtreinish:
@nonhermitian ok I've moved it to the advanced transpiler notebook

nonhermitian:
Is it possible to reduce the logging here.  That is a ton of logging.  Maybe just output timings or something that is not a verbose as the current example.

mtreinish:
@nonhermitian Oh, sorry missed your comment here. Yeah I can do that, it's because the example circuits in this notebook are larger than where i copied this from. I'll just do setup in this section to reduce the size of things we log.

mtreinish:
@nonhermitian let me know if I should reduce the amount of logging any more, I can run things with an even smaller circuit and a smaller pass manager to reduce it even more

nonhermitian:
@mtreinish This is much better.  Many thanks.

PR #786, opened 2019-08-31
fixing binder, again

omarcostahamido:


nonhermitian:
I think we are foregoing the binder in lieu if the fact that the tutorials are hosted on the IQX (hence the repo name change).

omarcostahamido:
[this](https://quantum-computing.ibm.com/jupyter), yes.
oh well, you can reject the pull, but it would be better to change the readme then.
Question: for the purpose of the self hosted notebooks on IQX, do you know if it needs the environment.yml?

O

jaygambetta:
changed the readme -- thanks for pointing this out

PR #840, opened 2019-12-10
Add job manager tutorial

jyu00:
### Summary
Add Job Manager tutorial to `3_the_ibmq_account` now that qiskit 0.14 is released.


nonhermitian:
I think this looks good.  However, is there an example of how to retrieve these multiple job objects and reconstruct the results?  You cannot just add two result objects together.

jyu00:
`job_set.results()` actually returns a `ManagedResults` object, which aggregates results for you. I clarified that and added an example on getting individual jobs in a set.

nonhermitian:
Yes, but my question deals with retrieving results from past jobs.  As far as I understand, you need to get both jobs, `5dcb153ef2b4ea00116071ba` and `5dcb153e1896660011bb286f` separately when retrieving past results.  However, this brings up a consistency issue.

For the combined managed job you can index collectively:
```python
results = job_set_bar.results()
print("This is the histogram data for experiment 5: {}".format(results.get_counts(5)))
print("This is the histogram data for experiment 15: {}".format(results.get_counts(15)))
```
But when with the two separate results you retrieved this same indexing does not work, e.g. there is no `15` index, and you will get an error.  So how does one retrieve results so that the live results and stored results are treated in the same way?  Can you do this, and if so can you give an example, or does this not work?

jyu00:
I see what you mean now. Unfortunately it does not yet support retrieving results from past jobs into a single `ManagedResults`. That feature would require api support to have some kind of identifiers that link the jobs together. I'll mention that in the tutorial.

nonhermitian:
Can you just replace the new master changes with your notebook changes?

PR #865, opened 2020-01-17
Add getting started with pulse tutorial

lcapelluto:
### Summary

An introductory tutorial for Pulse. This should wait until https://github.com/Qiskit/qiskit-terra/issues/3480 is done, so cell 6 can be updated with that functionality. It can then be released in sync with the next Qiskit release.

### Details and comments

There is still an issue of linking to the API Reference. As we build up the API Reference, it will be a very valuable tool for users. Especially now, when there are no further documentation tutorials to link to, the API Reference is even more valuable to cross reference, pointing users to where they should look next.

For instance, to link to the `Schedule` class, the syntax here is:

`<a href="https://qiskit.org/documentation/api/qiskit.circuit.QuantumCircuit.html" style="text-decoration:none"><code style="font-weight:bold;color:#a51920;">QuantumCircuit</code></a>`

whereas in sphinx, it would be:
`py:class:'qiskit.pulse.Schedule'` (where ' is a backtick)

Using a link rather than internal reference also slows down load time between pages. I wrote out the link in full for the first one, only, the second with the link but no style, and the rest without cross referencing.


lcapelluto:
On hold for Qiskit/qiskit-terra#3480

nonhermitian:
While this is waiting, you may want to take the chance to embed the header, make a TOC, and add the version table and copyright to the end.

nkanazawa1989:
Very good tutorial to start qiskit-pulse! I have some comments to improve

- unify the code style
In `Cell[3]` you added schedule without syntax sugar while `Cell[6]` uses the sugar. Not to confuse the beginners, `Cell[6]` is better to be written without sugar: `sched.insert(sched.duration, measure_all(backend))`. Maybe it's worth writing the sugar in separate cell and show both will give the same schedule plot.

- show label
You specified the name for `my_pulse` but it is not shown in the plot. I prefer to use `label=True` option with the drawer.

- intro
Not only error mitigation and characterization but also custom pulse design is good candidate of application. For the people not familiar with those application, it's better to add references to the articles.

nkanazawa1989:
Ah, and one more point. The title is `OpenPulse` or `Qiskit-Pulse`?

lcapelluto:
@nkanazawa1989 Thank you, that is great feedback! What is a use case for pulse design that doesn't fall under error mitigation and characterization? Good call to add articles, but the topics are so broad, I wouldn't know what to recommend. Maybe we can leave that to the textbook, instead, for the time being. Not sure what to do about the title :/

nkanazawa1989:
For example, we can create a custom entangular, CRz(theta) for example. Pulse level optimization such as GRAPE is also what we cannot do without pulse. I think @eggerdj can suggest good examples as an intro.

eggerdj:
We could start thinking on how to build pulse optimization functionality in Qiskit. This would probably mean implementing model based open-loop pulse shaping in Aer and closed-loop in Terra/Ignis perhaps(?). Note, however, that closed-loop tends to be very data intensive. Currently, this requires and RFC more than a notebook. I'll try and draft something up.

lcapelluto:
Hey guys, interesting ideas, but I don't understand how it relates to this PR. I'd be happy to include a "read more on theory" section at the end if you guys want to compile a list of recommendations, but that might warrant a new "discussion" issue, where we can continue thoughts there, and follow up with a PR after this is merged.

Regarding the first point, I think pulse optimization can fall under error mitigation. Assuming high performance/low error is what we are optimizing for, then as far as I can tell, it's two ways to say the same thing.

nonhermitian:
There is still no TOC.  See for example: https://github.com/Qiskit/qiskit-iqx-tutorials/blob/master/qiskit/fundamentals/0_start_coding_with_qiskit.ipynb

Also, some of the links point to 404.

lcapelluto:
@nonhermitian I removed all the links to the API reference docs. It's really too bad that we worked so hard to improve the API reference and have no way to link to them from these documentation pages.

lcapelluto:
For a getting started page, especially one as short as this, I think it would be better if the users didn't jump around when reading

lcapelluto:
I could use a mock backend (FakeAlmaden) but then we can't execute. I couldn't figure out how to get a working simulator in a few lines

PR #867, opened 2020-01-24
Add documentation for building Schedules using Pulse

lcapelluto:
### Summary

Explain:

- insert and `|`
- shift and `<<`
- append and `+`
- using the methods on both instructions and schedules
- common pitfalls


lcapelluto:
@nonhermitian I'll continue removing links to the API reference from all of these pages. Thank you for noticing that they had broken

lcapelluto:
No backends here 👍

PR #868, opened 2020-01-27
Pulse simulator tutorial using Duffing oscillator models

DanPuzzuoli:
### Update Feb 4:

I'm now happy with the current status of the tutorial, and am looking for comments!

### Summary

A tutorial demonstrating the `PulseSimulator` use-case of the user defining their own model, then running pulse simulations on it.

This is a work in progress; comments welcome! 

### Details and comments

Steps in the tutorial:
- A 2 qubit model using a Duffing oscillator model is defined
- A calibration routine is run to calibrate a pi pulse on each qubit (ripped from Dave's original example notebook)
- Using the calibrated pi pulse, state-dependent CR oscillations are demonstrated (it is shown using the bloch sphere that trajectory of the target qubit changes depending on the state of the driven qubit)


lcapelluto:
Awesome! I'd recommend renaming the tutorials to describe what they do, and make this into two separate PRs so that the one that is more ready doesn't have to wait on second one (and it is kinder to the reviewers)

aasfaw:
This might be a great fit for the textbook. What do you think, @DanPuzzuoli and @nonhermitian? It seems to work well with the theme of pedagogically presenting something and using qiskit to do it.

DanPuzzuoli:
I leave it up to you guys to decide. What are the standards for the textbook vs a tutorial; would anything need to be done to adapt it? @zachschoenfeld33 has been thinking about pulse sim textbook stuff as well.

nkanazawa1989:
Looks great! This is very useful notebook for hands-on events and lectures.
I have some comments

- model description
  - is there any reason not to use capital `J` for the coupling constant? it may look like imaginary number, especially for electronic engineers. also there is inconsistency of terms in the model description and the code, eg _frequency nu_ vs _oscillator_freqs_. they should be unified. 
  - it is better to write full two qubit Hamiltonian after Cell4 because the control channel doesn't appear in the model, but we will use it in the same notebook. also it is better to write relationship of their notations in the model (D, U, M) and pulse channel representations (pulse.DriveChannel, pulse.ControlChannel, pulse.MeasureChannel). I know measurement channel is not actually simulated.
  - as a simulator tutorial, it may be better to briefly explain how the simulator calculate the dynamics.

- pulse part
  - why you don't show pulse sequence? (ie `sched[0].draw()`?)

below is not related to the notebook but just from curiosity.
- why `dt` is stored in `PulseSystemModel` class? it is more natural to be in the qobj configuration.
- what is the reason of CR Rabi rate dependency on control qubit state? I guess there is strong IX interaction induced by higher levels and it is interesting to run echo sequence to remove this (this also shows the validity of physical model).

lcapelluto:
1.  Should provide some kind of definition or introduction to what a duffing osc model is in the first paragraph
 2 . This is also a good place to motivate the usage of the model. There is a sentence there listing "duffing model, example 1, example 2". As far as I understand, the model allows you to do the following two examples. You have to specify _some_ kind of model for qubit behavior to start doing any simulations of pulse-level qubit behavior, right? This relationship should be explained right away. What is the relationship between example 1 and 2? Can a reader skip to example 2 if they want, or is example 2 built on example 1?

#### Duffing system model

1. Think about swapping how you present the info: if you start with the physics, then the reader has the topic motivation and knows what they are working towards. Then, when you explain that it fits into the simulator as a `PulseSystemModel` with the convenient `duffing_system_model` function, they understand why they want that. This also leads you right into the next code block, where `duffing_system_model` is used. 
2. You could consider waiting to mention and explain `PulseSystemModel` altogether after the code block. They don't really need to know about it until they have just made one, as extra context.
3. `# time` comment over `dt` should be a little more specific ("timestep" or "sample duration" or "timescale")

#### Cross-resonance `ControlChannel` indices

1. first sentence: `the object` (I'd wonder, what object?) should be replaced with `two_qubit_model`
2. `E.g.` ... is providing an example of what is stated in the first sentence; should it follow that sentence rather than beginning a new paragraph?

#### Example 1

1. Would you mind replacing `pulse_lib.X` with the parametric pulses? This isn't that important, but I think it could be nice.
2. I like your `qubit_populations_from_memory` function! I've been thinking for a while about looking into the results object and adding some utility functions like this to terra

#### Example 2
1. `## Drive qubit 1 to observe CR oscillations on qubit 0` this is part of example 2, right? It should be a subheader of Ex 2, not the same level (`###`)

DanPuzzuoli:
Thanks for the comments! 

@nkanazawa1989 
> * model description
>   
>   * is there any reason not to use capital `J` for the coupling constant? it may look like imaginary number, especially for electronic engineers. also there is inconsistency of terms in the model description and the code, eg _frequency nu_ vs _oscillator_freqs_. they should be unified.

There is no reason for using `j` over `J` 😄 . In the string representation being used behind the scenes it is necessary to use lower case letters for coefficients, and I just carried that over when writing this. If it is standard physics notation I can change it to `J` (I gather that it is).

For the nu/oscillator_freqs, I think I want to keep it as is. In the simulator code the argument name is already set to `oscillator_freqs`.

>   * it is better to write full two qubit Hamiltonian after Cell4 because the control channel doesn't appear in the model, but we will use it in the same notebook. also it is better to write relationship of their notations in the model (D, U, M) and pulse channel representations (pulse.DriveChannel, pulse.ControlChannel, pulse.MeasureChannel). I know measurement channel is not actually simulated.

I think I may actually move the Hamiltonian description all the way up to the beginning of the notebook, as per @lcapelluto's comments. I like the idea of motivating the physics in a somewhat self-contained way at the beginning before getting into the code.

I agree that the channel naming/usage could be made more clear, especially in relation to the control channels. I'll try to do this in the next round of changes.

>   * as a simulator tutorial, it may be better to briefly explain how the simulator calculate the dynamics.

I'll try to think of where it is natural to do this.

> * pulse part
>   
>   * why you don't show pulse sequence? (ie `sched[0].draw()`?)
> 

👍 I'll add a cell drawing one of the schedules.

> below is not related to the notebook but just from curiosity.
> 
> * why `dt` is stored in `PulseSystemModel` class? it is more natural to be in the qobj configuration.

I agree to some extent; though I think in this particular context it is natural. If `PulseSystemModel` is meant to be the 'simulation/model version' of a real pulse backend, then it does make sense for `dt` to be stored here, as this is a property of the system as a whole (which includes the control electronics, which do have a fixed sample rate).

> * what is the reason of CR Rabi rate dependency on control qubit state? I guess there is strong IX interaction induced by higher levels and it is interesting to run echo sequence to remove this (this also shows the validity of physical model).

To be honest I haven't put the time into understanding the physics of the CR driving, so I can't answer that 🤣 . I agree that further exploration of this would be cool in subsequent tutorials (or a more expanded version of this one).

@lcapelluto 
> #### Intro comments and ####Duffing system model comment 1

I agree with these three comments; I'll move some pieces (like the Hamiltonian) to the intro, and expand the motivation/description of what's going to happen in the rest of the notebook.

> 2. You could consider waiting to mention and explain `PulseSystemModel` altogether after the code block. They don't really need to know about it until they have just made one, as extra context.

Makes sense.

> 3. Could the `# time` comment over `dt` be a little more specific? ("timestep" or "sample duration" or "timescale"?)
> 

Is the lone word "time" not enough of an explanation 😆 ?? Will change to "sample duration" or something along these lines.

> #### Cross-resonance `ControlChannel` indices
> 1. first sentence: `the object` (I'd wonder, what object?) should be replaced with `two_qubit_model`

👍 

> 2. `E.g.` ... is providing an example of what is stated in the first sentence; should it follow that sentence rather than beginning a new paragraph?
> 

👍 

> #### Example 1
> 1. Would you mind replacing `pulse_lib.X` with the parametric pulses? This isn't that important, but I think it could be nice.

I'll try this but I'm not sure what will happen. The code in the pulse simulator that interprets pulse schedules is unaware of parametric pulses, so it won't work unless they can be naively interchanged with non-parametric pulses. 

> 2. I like your `qubit_populations_from_memory` function! I've been thinking for a while about looking into the results object and adding some utility functions like this to terra
> 

It wasn't originally in a function, but this code (along with most of the notebook) was written by @dcmckayibm  :P. 

> #### Example 2
> 1. `## Drive qubit 1 to observe CR oscillations on qubit 0` this is part of example 2, right? It should be a subheader of Ex 2, not the same level (`###`)

👍

lcapelluto:
About parametric pulses: yes they will be assembled to sample pulses since the backend doesn't support it... but it's also fine to leave it as is!

lcapelluto:
Screen Shot 2020-02-05 at 3 49 14 PM
offset from max amp?

DanPuzzuoli:
@lcapelluto @nkanazawa1989 

Thanks again for your comments, I've just updated the tutorial and I think the pedagogy has improved. I've tried to address your comments; the biggest change by far is that I've expanded the intro to explain everything in the notebook at a "physics" level, with a brief description of how this will be performed in the code (as a consequence there is less description throughout the notebook).

@lcapelluto - I switched to parametric pulses, and somehow the issue you pointed out with the pi amplitude not aligning visually with the peak was fixed 😄 / 😢 ???

zachschoenfeld33:
I think it looks good! In terms of the textbook, I had done something very similar to you @DanPuzzuoli. The Rabi expt part is good for the textbook, but we were thinking of waiting until noise models are added so that we can fully replicate the current textbook chapter (including T1, T2, etc).

lcapelluto:
Love the changes! This tutorial is seriously awesome. Well written, very valuable new resource. :D 

Thank you so much @DanPuzzuoli !!

Some last comments and questions:
1. I know personally that it is a pain to do this for everything, but it's probably worth it to at least link `duffing_system_model` to the API reference docs. I've been using markdown style `[<backtick>text<backtick>](link)`. Can't find it in the API reference myself :(
2. Do people need to know about `zvode` for these docs?
3. I didn't realize until Ali mentioned in the meeting that the cutoff dimension is the dimension of the qubit state space. (I guess I thought it could be something more abstract, just the number of terms in some expansion to include, some arbitrary accuracy related value, like p in VQE? Idk) This seems obvious now -- I'm not sure if I was being dense or if it would help to clarify.
4. When you run the power rabi on qubit 1, you don't have to specify the qubit freq. Was this the case for qubit 0, too? Is the situation special for qubit 1?
5. "The model generating function duffing_system_model, automatically sets up" I think remove the comma in the middle or add one after 'function'

zachschoenfeld33:
Yes great work @DanPuzzuoli! Few extra comments on the Duffing oscillator tutorial:

Introduction 
Section 1. Duffing oscillator model
- remove mentions of `zvode`/`Scipy`, don't think the user needs to know how things are solved; I like the mention of the frame. Perhaps add a reference to rotating frame/rotating wave approximation? (Agree w/ Lauren’s point 2)
- for Duffing model, I'd add the full Hamiltonian, ie `H=` and then discuss both terms. I'd also add a sum over all qubits at the front so the full system can be specified. 

Section 2. `pi` pulse calibration
- mention Bloch sphere in describing `pi` pulse
- say "larger drive amplitude" rather than "larger amplitude"
- I'd note IQ plane and level 1 data when saying "the measurement data is fit"

General:
- Maybe call `meas_sched`->`meas_acq_sched` since you’re doing both?
- Note on what drift means (control terms zero)
- The measure and acquire commands are the same for individual qubits and the CR. Perhaps define this pulse at the top as a constant for the entire notebook.

haggaila:
Hi Dan, fashionably late I'm also going over your tutorials - they are great and teach me a lot. If any, I'll send you some further thoughts separately.

DanPuzzuoli:
Latest changes:

- Added table of contents, version table, and copyright
- Relabelled section numbers
- Simplified Rabi experiment section using Ignis functions as suggested by @dcmckayibm 
- Renamed the tutorial to match numbered naming conventions

I'd like to wrap this up. Are there any other key/necessary changes, or is this good to go? @ajavadia @nonhermitian @dcmckayibm

PR #898, opened 2020-05-08
fix terra adv 1

nonhermitian:
### Summary
Somehow terra/advanced/1_* is no longer a valid notebook for some reason.


mtreinish:
This will likely fail because we need to set the label manually for the custom multiqubit instruction otherwise the mpl drawer will crash. At least until https://github.com/Qiskit/qiskit-terra/pull/4389 merges and is included in a release. That's what I was attempting to do with that line (but messed it up)

mtreinish:
ok, now I'm confused how this passed. It definitely failed with an error the first time around which is why I pushed terra#4389

PR #900, opened 2020-05-08
legacy tuts qiskit header fix for #899

### LINKED ISSUE #899: legacy tuts don't display header while rendering on github
title says it all.

omarcostahamido:
### Summary
this fixes #899 
changed header image into relative path

### Details and comments
this allows one to update header once and have it effect on all ipynb.
this also makes all ipynb a little lighter since they are all referencing the same image and not trying to attach it to each notebook.


nonhermitian:
Many thanks for this.  I thought they were all embedded images but I guess I was wrong.

PR #915, opened 2020-05-20
try parallel build of tutorials

nonhermitian:


nonhermitian:
This works locally well, but not in CI

PR #916, opened 2020-05-21
Fix deprecation warnings in transpiler tutorial

Cryoris:
### Summary

The notebook `4_transpiler_passes_and_passmanager` used some old code that was still emitting loads of deprecation warnings. Re-running fixed a lot and updating a few lines of code fixed the rest.


Cryoris:
Didn't see #910 already addresses this.

PR #924, opened 2020-06-18
Remove pprint_as_string from optimization

Cryoris:
### Summary

The `pprint_as_string` method will be removed from the optimization module.


molar-volume:
in this tutorial, we can also change all
print(mod.export_as_lp_string()) ---> print(mod)

can I do it?

molar-volume:
@Cryoris the method print_as_lp_string() does not work on docplex.mp.model or QuadraticProgram, although I was not removing it

I would skip the whole line
"You can display your optimization problem e.g. with the `print_as_lp_string` method."

Cryoris:
@molar-volume nice to see you're motivated! But we cannot use `print()` yet because the tutorials should work with the latest official release (the current `stable` branch on GitHub) and the `print()` method was only added recently to master.

Good catch with the `print_as_lp_string`, I removed that line.

Cryoris:
The build fails due to a cell timeout, but all cells run in a few seconds. @nonhermitian do you know what's going on here?

nonhermitian:
Yes: https://github.com/Qiskit/qiskit-tutorials/pull/925

nonhermitian:
The file was run using an odd kernel name (seen this before):

```
No such kernel named python37564bitlatestvirtualenv76cfe9d7d3f749489cf2ff69593234af
```

The kernel name should be `python3`.

Cryoris:
Got it, thanks!

PR #925, opened 2020-06-18
Fix timeout error by moving to Py37

nonhermitian:
### Summary
As discovered by @mtreinish, Py 3.8.3 breaks `nbclient` in the same manner as:

https://github.com/tornadoweb/tornado/issues/2753


mtreinish:
I've opened https://github.com/ipython/ipykernel/issues/519 to report this issue upstream

PR #962, opened 2020-07-15
cleanup finance latex in rst

nonhermitian:
### Summary
When converting from notebooks to rst, there always needs to be a space around the math sections otherwise they do not render.


nonhermitian:
ehh, needs more work.

PR #969, opened 2020-07-21
pin mpl to less than 3.3 till new Qiskit

nonhermitian:
### Summary
Fixes bug in Qiskit that is not fixed until new version is out.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -37,6 +37,7 @@ steps:
     virtualenv /tmp/docs_build
     source /tmp/docs_build/bin/activate
     pip install -U qiskit jupyter sphinx nbsphinx sphinx_rtd_theme
+    pip install matplotlib<3.3
     pip install -U qiskit-terra[visualization] cvxpy
     sudo apt-get install -y pandoc graphviz
   displayName: 'Install dependencies'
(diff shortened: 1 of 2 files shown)


mtreinish:
To fix this in CI you'll have to add matplotlib as an explicit requirement (and before qiskit) here: https://github.com/Qiskit/qiskit-tutorials/blob/master/azure-pipelines.yml#L40 Alternatively we probably should add a `constraints.txt` file and add `-c constraints.txt` to the pip command in CI so we don't have to hand modify the pip command line for a broken dependency.

nonhermitian:
Yeah your absolutely right.  I forgot that that yaml file was for quickly setting up users environments. Hopefully this is a short lived fix.

azure-pipelines.yml
```diff
@@ -37,6 +37,7 @@ steps:
     virtualenv /tmp/docs_build
     source /tmp/docs_build/bin/activate
     pip install -U qiskit jupyter sphinx nbsphinx sphinx_rtd_theme
+    pip install matplotlib<3.3
```
mtreinish, line 40:
```suggestion
    pip install 'matplotlib<3.3'
```

PR #973, opened 2020-07-30
Working version of the old tomography tutorial

gadial:
### Summary
This PR returns to qiskit-tutorials the tomography tutorial that was once part of it, but was removed due to run failures.

### Details and comments
This version should work with the most up-to-date version of Qiskit, given a correct installation of `cvxpy`, which means it was correctly linked with `blas` and `lapack`; othewise running will result in an error about the SCS engine being unable to solve SDP's.


nonhermitian:
Seems like the original issue still occurs here as it is failing with a SCS error.

gadial:
> Seems like the original issue still occurs here as it is failing with a SCS error.

Unfortunately that's not the original issue (which was `state_fidelity` failing due to missing `validate=False` argument) but something new and much more troublesome which affects Ingis in general.

chriseclectic:
You need to install `cvxopt` in the doc build environment (@mtreinish)

nonhermitian:
There are some formatting issues in the rendered docs:

Screen Shot 2020-08-06 at 15 38 17


Screen Shot 2020-08-06 at 15 38 31


Also not sure about this:

Screen Shot 2020-08-06 at 15 39 38

gadial:
> There are some formatting issues in the rendered docs:

I fixed the rendering issue. As to the "FATAL: cannot solve SDP", sadly this is the expected behavior - currently Ignis performs a test run to check whether some engine can be used; if it cannot, it uses another, but in the process this "FATAL" message is sent. We need to fix this inside Ignis itself.

nonhermitian:
It seems that the PR includes changes to 21 different files.  Can this be trimmed down to include only the tomography tutorial.  Otherwise, a list of changes to the other files should be given as well.  Otherwise, the tomo one looks good to go.

gadial:
> It seems that the PR includes changes to 21 different files. Can this be trimmed down to include only the tomography tutorial. Otherwise, a list of changes to the other files should be given as well. Otherwise, the tomo one looks good to go.

This is due to some rebase problem I couldn't fix. I think I'll have to create a new PR with the fixed notebook.

PR #983, opened 2020-08-06
Fixed Finance data providers access

manoelmarques:
### Summary
Fix Provider tutorial errors and adds a new Yahoo! Provider


nonhermitian:
I believe this was fixed in #986.  Re-open otherwise.

woodsp-ibm:
The 0.19.6 code had a bug in it that was corrected in 0.20.0. However the tutorial had a related issue and now that things worked in 0.20.0 the old tutorial had an issue because of that. Since it was wanted to run the old tutorial on the new 0.20.0 the existing tutorial has the defective section removed. This already had a fix in it ready to go and would run with the new 0.20.0 but for the CI APi check the old one was required to be run. This should be opened/merged to fix things as they should be as well as show a new feature added for 0.20.0.

manoelmarques:
@nonhermitian Could you please re-open it ? It seems I don't have authorization.

nonhermitian:
It says the branch was force pushed or recreated and will not let me reopen as well.  Not sure what that actually means to be honest.  Maybe just make another PR

PR #984, opened 2020-08-07
removing backend properties as a dictionary

1ucian0:
`backend.properties` should be a `BackendProperties` class. The transpiler assumes that.


nonhermitian:
Closing in favor of #985 because there the full notebook was executed for constancy.

PR #986, opened 2020-08-07
Remove broken cells for aqua 0.7.4

mtreinish:
### Summary

In the latest Aqua release 0.7.4 several issues with external
data providers were fixed so that they properly raise errors when they
data providers were incorrectly used. The tutorial happened to be
working by chance  because it was never actually being used since it
requires a token from an external service. It was only passing because
of the lack of checking. This has been fixed in 0.7.4 but it now raises
an error. To fix the tutorials new APIs (which is technically a violation
of the stable branch policy [1]) but this is blocked on the metapackage
release. To enable releasing the new metpackage and building a working
tutorial the broken code in this tutorial can't exist in tree. This
commit removes the section of the tutorial which do no work, we can
investigate adding them back after the release.

### Details and comments

[1] https://qiskit.org/documentation/contributing_to_qiskit.html#stable-branch-policy


mtreinish:
This is needed to unblock Qiskit/qiskit#998

PR #1001, opened 2020-08-11
Fix failing CI

lcapelluto:
### Summary

Hopefully this fixes the issue that is blocking my other PR.

Add optional pip install for pyscf which is required for qiskit chemistry

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -39,6 +39,7 @@ steps:
     pip install -U qiskit jupyter sphinx nbsphinx sphinx_rtd_theme
     pip install 'matplotlib<3.3'
     pip install -U qiskit[visualization] cvxpy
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -39,6 +39,7 @@ steps:
     pip install -U qiskit jupyter sphinx nbsphinx sphinx_rtd_theme
     pip install 'matplotlib<3.3'
     pip install -U qiskit[visualization] cvxpy
+    pip install pyscf
```
mtreinish, line 42: Let see what CI says, but in other envs we've had to cap this as

```suggestion
    pip install 'pyscf<1.7.4'
```

because pyscf added a version cap on scipy which conflicts with what everything else installs

PR #1020, opened 2020-08-14
'mpl'

shovitraj:
matplotlib inline added in the circuit for vizualization


shovitraj:
'mpl' added to circuit drawing for better vizualization.

nonhermitian:
Not sure if this is a desired change.  The tutorial is supposed to go with the default options until the point where it mentions the `mpl` drawer.

PR #1030, opened 2020-08-28
Fix Aqua deprecated messages

manoelmarques:
### Details and comments

It removes deprecated messages from the notebooks by updating them to the latest Aqua API on Pypi.
It is related to Qiskit/qiskit-aqua#1220

Probably the fix in the finance notebooks should be added to Aqua MultivariateProblem class. @stefan-woerner could you please look at this change ?


tutorials/finance/07_asian_barrier_spread_pricing.ipynb
```diff
@@ -57,11 +57,32 @@
...
     "from qiskit.aqua.algorithms import IterativeAmplitudeEstimation\n",
-    "from qiskit.aqua.circuits import WeightedSumOperator, FixedValueComparator as Comparator\n",
+    "from qiskit.aqua.circuits import WeightedSumOperator\n",
     "from qiskit.aqua.components.uncertainty_problems import UnivariatePiecewiseLinearObjective as PwlObjective\n",
     "from qiskit.aqua.components.uncertainty_problems import MultivariateProblem\n",
-    "from qiskit.aqua.components.uncertainty_models import MultivariateLogNormalDistribution"
+    "from qiskit.aqua.components.uncertainty_models import MultivariateLogNormalDistribution\n",
+    "\n",
+    "\n",
+    "class Comparator(IntegerComparator):\n",
```
mtreinish, line 68: The only thing here (and the other tutorials with this class) I think is that we should have a comment on what the purpose of this class is. It's not exactly clear when reading this tutorial why we need to define this class. We're instructing people that they need to write this thing (at least for now) so it would be good to just mention why.

mtreinish (approved):
Thanks for adding the comment. LGTM

tutorials/finance/07_asian_barrier_spread_pricing.ipynb
manoelmarques: comments added.

PR #1037, opened 2020-09-17
Adjust sections in chemistry notebook

1ucian0:
The finance sections look like this:
Screen Shot 2020-09-17 at 4 35 17 PM

The chemistry section looks like this:
Screen Shot 2020-09-17 at 4 34 51 PM

None of the other sections have a "contributors" sections from where the other subsections hang. So this PR adjust it to make it more similar to the other sections. Also, I'm removing the "The latest version of this notebook is..." note.


nonhermitian (approved):
Thanks for spotting this.

PR #1051, opened 2020-10-15
replacing big endian with little endian concept for bit ordering in q…

michalkordyzon:
…iskit

Please verify me if I got it correct. I think qiskit uses little endian bits and qubits ordering with LSB coming first. This documentation claims that qiskit uses big endian with MSB coming first, and in the next sentence it shows little endian: |10>  where 0 is state of first qubit and 1 the state of second qubit.


nonhermitian:
The LSB is the right most bit in the bit strings.

ertogrul:
> The LSB is the right most bit in the bit strings.

nonhermitian, agreed. Is the change I am proposing correct?

ertogrul:
ahhh, I can see that, I have mistaken left and right in the text of documentation. Let me withdraw this pull request and prepare the correct one. Thank you, nonhermitian.

divshacker:
Are you still working on this @ertogrul

PR #1067, opened 2020-10-24
Add new line for list to be rendered properly

azeemba:
### Summary

The two lists are not being rendered as lists so we add an extra newline for it to behave properly.

### Details and comments

I have not added an item in the changelog since it is a minor fix. Let me know if it is preferred that I updated the changelog as well.

This is how it currently looks ([link](https://qiskit.org/documentation/tutorials/circuits/3_summary_of_quantum_operations.html)): 

image

This is how it looks after the fix:

image


SooluThomas (approved):
LGTM!

PR #1072, opened 2020-11-03
Fix incorrect fidelity calculations in tomography tutorial

chriseclectic:
### Summary

* Fixes errors in the process and gateset tomography examples because of incorrect usage of the `process_fidelity` function. 
* Updates examples to get the target unitary / statevectors using the quantum info classes rather than the unitary / statevector simulator backends.

### Details and comments

The use of the `process_fidelity` function was incorrect for multiple reasons:

* The function must be passed channel objects, not numpy arrays otherwise it is treating the input matrices as unitaries
* The first argument must be the noisy channel and the second argument the unitary channel
* Disabling checking for `require_cp` prevented the errors that would have been generated from the incorrect inputs being shown (this will require a separate patch in Terra but cant be done until the tutorial is updated or it wont pass CI).

This PR removes the manual disabling of input validation for the fidelity functions and changes the fitter method to be `lstsq` to avoid issues with the poor performance of the SCS solver in CVPXY if the CVXOPT solver is not found on the system. In general the fits should not require disabling validation unless something is wrong with the solver / data. Unfortunately the default SDP solver (SCS) in CVXPY has poor accuracy and fails to respect CP constraints. I plan to remove it from the auto method in a patch to ignis.


nonhermitian (changes requested):
Just needs the header image removed then should be good.

chriseclectic:
@nonhermitian done

nonhermitian (approved):

PR #1068, opened 2020-10-27
Add Aqua operators tutorial

jlapeyre:

PR #1077, opened 2020-11-06
Intro tutorial on Aqua operators

hay-k:
### Summary
This is a basic introductory tutorial covering the main functionalities in Aqua's operator flow layer. Contains introduction to all major components (State functions, operators, evolutions, measurements and expectation), as well as a simple end-to-end example demonstrating the usage of all this components in unison.

### Details and comments
@e-eight and myself have created this tutorial and there are plans to create more advanced ones as well, but since we will be busy with the Qiskit challenge 2020 this month, we thought it is worth submitting this one, and work on the rest later.


t-imamichi:
Please take a look at #1068

tutorials/operators/01_getting_started_with_operators.ipynb
```diff
@@ -0,0 +1,951 @@
...
+   "source": [
+    "# Getting started with operators\n",
+    "\n",
+    "Qiskit provides sophisticated functionality with operators: [The \"Operator Flow\" layer in Aqua](https://qiskit.org/documentation/apidoc/qiskit.aqua.operators.html). In this tutorial notebook we will get familiar with the main building blocks of the operator flow layer:\n",
+    "\n",
+    "1. State functions\n",
+    "2. Operators and evolutions\n",
+    "3. Measurements and expectations\n",
+    "\n",
+    "This tutorial assumes basic familiarty with qubits, quantum superposition and quantum circuits.\n",
```
jlapeyre, line 15: *familiarty* misspelled.

jlapeyre:
Thanks very much for this contribution! There is obviously some overlap with #1068, as well as good material unique to each of the two tutorials. To name one particular, I think the additional detail in the explanation of of `PauliExpectation` in #1077 is important.

We could unify these, or alternatively, make them more complementary.

I have one high-level comment. I think that users would find understanding aqua operators much easier if we were to use standard terms and concepts, such as those found in Nielsen and Chuang, rather than those found in the Watrous book (which I personally like!) And use those terms precisely and consistently. This is what I tried to do in #1068. Let me explain with a few examples.

* If we talk about state vectors and linear operators in the standard way, then we don't need to explain what they mean, because it is understood. For example, in #1068 I talk about how to get a matrix element from an operator by specifying two indicies. This requires no additional explanation because this is how people are used to thinking about operators. For example, there is no need to discuss in the abstract different ways to view the Pauli `X` operator (with no reference to code).

* The definition of a measurement in Watrous is equivalent to the definition of a POVM in Nielsen and Chuang. But, this is not what the word "measurement" means in the operator documentation. Furthermore, Nielsen and Chuang define observables as a sum over projectors times outcomes. It seems like the latter is the way measurements are specified in the aqua operators, as observables. For example `StateFn(Z).adjoint()` represents the Pauli `Z` operator interpreted as an observable. It's important to consistently use the precise term *observable* here. Furthermore, we should avoid any confusion of inner products and observables (or measurements). This is important because, as an implementation detail, they share some of the same programming language objects.  For example `(~One).compose(One)` represents the inner product <1|1>, or one might say, the product of a bra and a ket.

* When possible, each expression given in code should be accompanied by the corresponding mathematical expression in standard notation. This aids immensely in understanding the code. I did that in most places in #1068. But, not everywhere, for example in input box 42. This should be fixed!  Hybrid code/math or adhoc notation should be avoided. For example `f(Zero) = One`.

When I look at #1077, it seems there is some tendency to follow some of the advice above. I would encourage authors to embrace it whole heartedly when writing tutorials.

e-eight:
@jlapeyre I like the idea of following Nielsen & Chuang, instead of Watrous for the tutorials, even though I too like Watrous a lot. We can definitely update #1077 to bring it more in line with Nielsen & Chuang. But since there is significant overlap between #1068 and #1077, and both are attempts at making the first tutorial on Operators, I think it might be better to unify the overlapping parts in to a single tutorial, and the leftovers can be complementary tutorials.

For example we can put the Pauli operators, state functions, and measurements in the first introductory tutorial, because those are topics that everyone who is going to use Aqua will probably know. And then we can have a following tutorial where we discuss evolution, and an end-to-end example.

PR #1091, opened 2020-11-25
Bump cell timeout in sphinx config

mtreinish:
### Summary

The `07_asian_barrier_spread_pricing` tutorial is quite slow and often
times out in CI because running the simulation takes very close to 3
minutes in the fastest case. But, the performance of the CI nodes is
quite variable since they run on virtual machines in azure's public
cloud. So many jobs fail randomly because of timeouts on that notebook.
This commit attempts to fix this by bumping the timeout 60 seconds which
hopefully will give us enough slack time in the job to handle slow CI
nodes.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -73,7 +73,7 @@ html_sourcelink_suffix = ''
 exclude_patterns = ['*.ipynb', '_build', 'legacy_tutorials',
                     '**.ipynb_checkpoints']
 
-nbsphinx_timeout = 180
+nbsphinx_timeout = 240
 nbsphinx_execute = 'always'
 nbsphinx_execute_arguments = [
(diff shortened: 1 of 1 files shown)


nonhermitian:
So I was wondering if we can keep the limits here but use a env var in meta to bump it up there. My worry is that if we bump it up now, there will be something in the future that requires another bump, etc etc

mtreinish:
That's a good idea I'll make it an env variable flag that defaults to 3 min. The issue is more than just the metapackage we've been hitting this a lot on terra's CI too. But having an env var lets us keep things strict here but have a relief valve for downstream ci.

nonhermitian (approved):

PR #1098, opened 2020-12-07
add widget notebook

abbycross:
### Summary

Composer Widget notebook

### Details and comments

I named it 1a so that it falls after the first notebook in the Circuits folder. Is there a better place for it?

### DIFF
--- a/tutorials/circuits/1a_Composer_widget.ipynb
+++ b/tutorials/circuits/1a_Composer_widget.ipynb
@@ -1,0 +1,1538 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Circuit Composer and Visualizer widgets"
+   ]
+  },
+  {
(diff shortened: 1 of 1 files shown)


abbycross:
FYI @beckykd

nonhermitian:
The tutorials cannot access the IQX as we want to minimize any blockers due to API issues.

abbycross:
@nonhermitian Better to turn this into a purely text document and post in the doc repo, or to change it to pure markdown within this current file so that it doesn't call IQX, and upload it here?

nonhermitian:
Let me play with this a bit first to see what happens when I try to build it into some docs.

raulotaolea:
@abbycross @nonhermitian The IQX widget library comes pre-installed in the IQX Jupyter Notebook, but it is not available in a public package manager yet. So notebooks outside IQX will not be able to use it.

nonhermitian:
Ok, then will close this for now till the time comes.

PR #1135, opened 2021-02-27
pin jax versions to load cvxpy

nonhermitian:
See: https://github.com/Qiskit/qiskit-aqua/pull/1524

Looks like some installs break the import of others.  Try the fix from the above PR


nonhermitian:
This can't fix the problem because jax is not even installed here.

PR #1141, opened 2021-03-09
Adding tutorial of GPU simulator for Qiskit Aer

doichanj:
### Summary
I added simple examples to show how to use GPUs to simulate quantum circuits on Qiskit Aer.

### Details and comments
This document is valid for Qiskit Aer 0.8 or later.


divshacker (approved):

nonhermitian:
I am curious how this tutorial helps to use the GPU simulator given that the tutorial just goes and runs on a CPU simulator.  It is also a bit odd that the gpu setting is hidden within backend options.  Is the GPU simulator not its own simulator instance?

doichanj:
Because the system of CI does not have GPUs, the test fails if tutorial only has GPU code. So I included CPU code for non-GPU environment by checking availability of methods in the simulator.

chriseclectic:
I think this tutorial needs a bit more work and should be renamed to something like *HPC Cluster Simulation*

@nonhermitian This tutorial should be for advanced HPC usage and describes how to set options for running on a HPC server with multiple GPUs. Instructions for basic GPU simulation on a single GPU can be covered in the Aer Provider tutorial that just requires selecting the gpu backend from provider (if its available) and running as normal.

@doichanj We need to update this now 0.8 is released to use the `AerSimulator` instead of the `QasmSimulator`. I think we should also include a section on how one would set simulator options to run a large MPI simulation on a properly configured cluster.

Eric-Arellano:
Aer tutorials should now live in the Aer documentation at https://github.com/Qiskit/qiskit-aer. So closing this one, but consider re-opening in their repository.

Pardon that this didn't get a more prompt review and thank you for making Qiskit better!

PR #1162, opened 2021-04-07
Remove Aqua references from notebooks

manoelmarques:
### Summary

Aqua imports replaced by QIskit Terra classes.
Chemistry, Finance, Machine Learning and Optimization were removed as all those were converted and migrated to the new application repositories


tutorials/algorithms/02_vqe_convergence.ipynb
```diff
@@ -271,10 +276,8 @@
    "outputs": [],
    "source": [
     "import logging\n",
-    "from qiskit.aqua import set_qiskit_aqua_logging\n",
-    "\n",
-    "\n",
-    "set_qiskit_aqua_logging(logging.INFO)"
+    "logging.basicConfig(level=logging.INFO)\n",
+    "logging.getLogger('qiskit.algorithms').setLevel(logging.INFO)"
```
woodsp-ibm, line 280: Since we could not really run and show logging I had just copy/paste a couple of lines from an actual run into the notebook in the past. I think the lines are still valid as such but we should edit out the `.aqua` in them which would make them correct for the current code.

tutorials/algorithms/04_vqe_advanced.ipynb
```diff
@@ -159,7 +171,7 @@
    "source": [
     "Here we see that result was arrived at much more quickly where `optimizer_evals` is 10 versus 72 when it started from a random value when the initial point was not supplied (default of None).\n",
```
woodsp-ibm, line 172: > Here we see that result was arrived at much more quickly where `optimizer_evals` is 10 versus 72

the `10 versus 72` needs updating - it looks like `9 versus 65` now. I did look to see whether it was possible to substitute values derived from the code into the text in the past as I wondered if things might change, but that did not seem readily doable. Maybe since it has changed, and maybe will do again, perhaps we should print some message via the code to highlight the different rather than having this in a text section.

tutorials/algorithms/04_vqe_advanced.ipynb
```diff
@@ -184,37 +196,41 @@
...
     "\n",
     "ansatz = TwoLocal(rotation_blocks='ry', entanglement_blocks='cz')\n",
     "slsqp = SLSQP(maxiter=1000)\n",
-    "vqe = VQE(operator=H2_op, var_form=ansatz, optimizer=slsqp, quantum_instance=qi, include_custom=True)\n",
-    "result = vqe.run()\n",
+    "vqe = VQE(ansatz, optimizer=slsqp, quantum_instance=qi, include_custom=True)\n",
+    "result = vqe.compute_minimum_eigenvalue(operator=H2_op)\n",
     "\n",
-    "pp.pprint(result)"
+    "pp.pprint(str(result))"
```
woodsp-ibm, line 233: In the text below this- again it includes values from the results that look like its needs updating

> The optimal value can be seen to be wrong too, i.e. -1.098 versus the correct -1.857.

The first value, the wrong one, looks now to be -1.074

tutorials/algorithms/09_textbook_algorithms.ipynb
```diff
@@ -4,13 +4,11 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Textbook and Shor's algorithms\n",
+    "# Shor's algorithms\n",
     "\n",
-    "Qiskit contains implementations of the well known textbook quantum algorithms such as the [Deutsch-Jozsa algorithm](https://qiskit.org/textbook/ch-algorithms/deutsch-jozsa.html), the [Bernstein-Vazirani algorithm](https://qiskit.org/textbook/ch-algorithms/bernstein-vazirani.html) and [Simon's algorithm](https://qiskit.org/textbook/ch-algorithms/simon.html).\n",
+    "Qiski has an implementation of [Shor's algorithm](https://qiskit.org/textbook/ch-algorithms/shor.html).\n",
```
woodsp-ibm, line 9: Should we rename the notebook itself too - say `09_factorizers` since this is all this notebook is about now

woodsp-ibm (approved):

mtreinish (approved):
This LGTM, the only thing that sticks out is the use of BasicAer everywhere (since that's not really a change). We really shouldn't be using it in the tutorials because it's buggy and slow (it's only for places where aer isn't available) and we don't want to encourage it's use. But we can fix that in a follow up.

manoelmarques:
I can still change all to use Aer in this PR if nobody has any objections. It should be a simple change.

nonhermitian:
I think we need to leave this on hold for now, and wait until the next release to remove everything.  There is likely to be a non-trivial subset of users who either cannot or do not want to update to the new modules at this time, e.g. those in the middle of getting results / writing papers.  As such, these should all remain for now.

woodsp-ibm:
@nonhermitian You are advocating keeping the former Aqua set of tutorials, that are being deleted by this until the next release? The one thing we need to address is algorithms and operators, since, unlike the applications areas of optimization etc we do not have a copy (new updated set) elsewhere. We could rename algorithms and operators that are here to something like algorithms_legacy and operators_legacy and the new ones based on the current code in Terra are in the original place, i.e. algorithms and operators. Tutorials index would just have to be adjusted to have both sets with the legacy ones having the deprecation banner.

nonhermitian:
Indeed, I am saying we should keep the deprecated Aqua stuff around until the next major release.  The docs already say the relevant tutorials are deprecated, and one should look in the new modules for those items.

mtreinish:
@nonhermitian I think we can consider this unblocked now. We're hosting stable versions of the documentation including tutorials for previous releases on qiskit.org/documentation, for example: https://qiskit.org/documentation/stable/0.24/index.html for 0.24 before the aqua deprecation. (they're also accessible/discoverable from the dropdown on the main documentation page). So if people are on old versions they can just refer to the older release's documentation.

nonhermitian:
I think this would still be picked up in a point release would it not?

nonhermitian:
So if we fine with a point removing the tutorials from 0.26 then ok by me.

tutorials/operators/02_gradients_framework.ipynb
```diff
@@ -215,19 +215,11 @@
...
-      "Gradient [(1.414213562373095+1.12e-16j), 0j]\n"
-     ]
-    }
-   ],
+   "outputs": [],
    "source": [
     "# Assign the parameters and evaluate the gradient\n",
-    "grad_result = grad.assign_parameters(value_dict).eval()\n",
-    "print('Gradient', grad_result)"
+    "# grad_result = grad.assign_parameters(value_dict).eval()\n",
```
mtreinish, line 221: In talking to @Cryoris he suggested replacing `~StateFn` with `StateFn(..., is_measurement=True)` to avoid the failure this is hitting. If we do that we should be able to uncomment this line

manoelmarques: I made the change and it fixed the problem.

Cryoris:
I can reproduce the failures locally with 0.17 but I couldn't find a way to resolve it (like taking the conjugate of the coefficients) that works also with 0.18. I'd say just remove the coefficients and show how to compute the gradients with a fixed operator, i.e.
```
H = (2 * X) + Z
```

mtreinish (dismissed):
LGTM, but I'll wait for @Cryoris, @stefan-woerner, or @woodsp-ibm  to confirm everything looks good here before merging since they have more domain expertise.

One thing I think we should do after this merges is to add a redirect from the old docs hosted tutorials to the application docs hosted versions. This should be pretty easy and something we can probably do by just adding entries to the redirect links like:
```
'tutorials/optimization/*": "https://qiskit.org/documentation/optimization/tutorials/$source"
```
to https://github.com/Qiskit/qiskit/blob/master/docs/conf.py#L82-L84

(see https://documatt.gitlab.io/sphinx-reredirects/usage.html for more details)

peendebak:
@manoelmarques In the updated notebook `05_qaoa.ipynb` a different problem is solved than in the original notebook. QAOA is used for partitioning problem on a graph without taking the weights into account. The graph definition (and the corresponding plot) at the top if the notebook have weights in them. To avoid confusion I suggest to either i) update the notebook to solve the problem with weights, or ii) update the definition of the graph to have all edges weight 1.

peendebak:
@manoelmarques Also from `05_qaoa.ipynb`. What is the goal of `get_graph_solution`? There is no explanation. Also if `x` is a solution, then `1-x` is a solution just as good as the objective function is invariant with respect to this operation.

tutorials/algorithms/03_vqe_simulation_with_noise.ipynb
```diff
@@ -300,12 +300,2652 @@
     "scrolled": true
    },
    "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "/Users/manoel/projects/manoelmarques/qiskit-terra/qiskit/utils/measurement_error_mitigation.py:44: DeprecationWarning: Back-references to from Bit instances to their containing Registers have been deprecated. Instead, inspect Registers to find their contained Bits.\n",
```
Cryoris, line 307: Do we want to hide these deprecation warnings for the notebook? Or maybe update `measurement_error_mitigation`?

tutorials/algorithms/03_vqe_simulation_with_noise.ipynb
```diff
@@ -226,7 +226,7 @@
     "from qiskit.test.mock import FakeVigo\n",
     "device_backend = FakeVigo()\n",
     "\n",
-    "backend = Aer.get_backend('qasm_simulator')\n",
+    "backend = Aer.get_backend('aer_simulator')\n",
     "counts1 = []\n",
     "values1 = []\n",
     "noise_model = None\n",
```
Cryoris, line 232: Is disabling parallelism still necessary? If yes, do we want to add a comment? 🙂

tutorials/algorithms/05_qaoa.ipynb
```diff
@@ -0,0 +1,398 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 7,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "[1 1 0 0]\n",
```
Cryoris, line 285: This solution doesn't match the others, do we know why?

tutorials/algorithms/06_grover.ipynb
```diff
@@ -0,0 +1,746 @@
...
+    "    print(ex)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "tags": []
+   },
+   "source": [
+    "You can observe, that this oracle is actually implemented with three qubits instead of two!\n",
```
Cryoris: This should print the oracle, otherwise we don't see that it uses 3 qubits 😄

Cryoris (commented):
Overall looks good, only in the QAOA example it's strange that the NumPy result doesn't match the others. 

I left some other minor comments, but as these are general issues and not necessarily related to porting Aqua to Terra feel free to skip them 🙂

woodsp-ibm (approved):

tutorials/algorithms/03_vqe_simulation_with_noise.ipynb
manoelmarques: The deprecation msgs are eliminated in Qiskit/qiskit-terra#6503

tutorials/algorithms/03_vqe_simulation_with_noise.ipynb
manoelmarques: Removed statement

tutorials/algorithms/06_grover.ipynb
manoelmarques: Printing oracle now

woodsp-ibm (approved):

mtreinish (approved):

PR #1165, opened 2021-04-10
Updating description of a layer of operations.

3yakuya:
### Summary
Fixing layer description in "Implementing a BasicMapper Pass"

### Details and comments
Original sentence stated "each layer is a group of operations that does not act on independent qubits, so in theory all operations in a layer can be done independently", which is incorrect. Every layer is a group of operations specifically acting on separate qubits each. The new sentence clarifies that with "each layer is a group of operations acting on independent qubits..."


MaldoAlberto:
Pardon that this took us so long to review! Thank you for making Qiskit better!
Looks good to me.

Eric-Arellano (approved):

Issue #1172, opened 2021-04-23
save_unitary state error? It's not saving the unitary and providing an output.


WiFisunset:
### Information

- **Qiskit IBMQ Provider version**: 0.12.3
- **Python version**: 3.8.8
- **Operating system**: Windows 10

### What is the current behavior?
save.unitary creates an "AttributeError: 'QuantumCircuit' object has no attribute 'config' "

The same happens for the following commands:
result = simulator.run(circ).result()
unitary = result.get_unitary(circ)

### Steps to reproduce the problem
Following the tutorial code at (https://qiskit.org/documentation/tutorials/simulators/1_aer_provider.html): 

import numpy as np

from qiskit import QuantumCircuit
from qiskit import Aer, transpile
from qiskit.tools.visualization import plot_histogram, plot_state_city

circ = QuantumCircuit(2)
circ.h(0)
circ.cx(0, 1)
circ.save_unitary()

simulator = Aer.get_backend('aer_simulator')
circ = transpile(circ, simulator)

result = simulator.run(circ).result()
unitary = result.get_unitary(circ)
print("Circuit unitary:\n", unitary.round(5))

### What is the expected behavior?
The expected behavior is a printed out circuit unitary.

### Suggested solutions

Call save_unitary() after the Simulator has been defined.

AshwinKul28:
HI @WiFisunset The same code is working for me, 

Qiskit version: 0.17.1
OS: macOS Big Sur: 11.2.1
Python Version: 3.8.3

<img width="974" alt="Screenshot 2021-04-26 at 1 40 17 AM" src="https://user-images.githubusercontent.com/19169648/116008053-5d5dc700-a630-11eb-9ff5-3aa40553a186.png">

Maybe an update in qiskit version might be helpful for you.

WiFisunset:
Hi @AshwinKul28, I updated qiskit via 'pip install update qiskit==0.17.0', and I still got the error (with a different qiskit version than what was updated).
![get-unitary-error](https://user-images.githubusercontent.com/33338744/116039878-241d5b00-a620-11eb-9bf8-4fe616c69152.PNG)

WiFisunset:
I was able to fix the error, but only by having 'circ.save_unitary()' called After 'circ = transpile(circ, simulator)'. 

![get-unitary-error-fix](https://user-images.githubusercontent.com/33338744/116165266-57a7c600-a6b0-11eb-8f5c-3c0dd74150ce.PNG)

[...]

PR #1174, opened 2021-04-27
Update to 1_aer_provider.ipynb in the save_unitary program

AshwinKul28:
### Summary
The circ.save_unitary() call is used to save the simulator state as a unitary matrix of the run circuit. In the example numbered as [10] in the simulators/1_aer_simulator notebook, the circ_save_unitary() call has been made before simulator initialization. Hence it is failing to run when a user tries to run exact same program.

### Details and comments
Since, the save_unitary() function call described as follows mentions that save_unitary() function saves the state of the simulator. (references: [ref1](https://qiskit.org/documentation/stubs/qiskit.circuit.QuantumCircuit.html#qiskit.circuit.QuantumCircuit.save_unitary), [ref2](https://qiskit.org/documentation/tutorials/simulators/1_aer_provider.html))

Screenshot 2021-04-28 at 1 11 27 AM

But according to the program if we don't call the simulator before the circ.save_unitary call, we cannot save the state of the simulator and hence QuantumCircuit is giving an error as follows:

Screenshot 2021-04-28 at 1 03 30 AM

Hence I have given the fix in the document as I'm calling save_unitary() function after the simulator initialisation.

Screenshot 2021-04-28 at 1 04 03 AM

### Git Issue
https://github.com/Qiskit/qiskit-tutorials/issues/1172

PR #1176, opened 2021-05-03
Save states need 'from qiskit.providers.aer import Aer' called first in order to function.

WiFisunset:
### Summary
Save simulator states need to be called after the Simulator in order to function. Otherwise the tutorial will output a "AttributeError: 'QuantumCircuit' object has no attribute 'save_state'" error. This issue affects all save simulator states in the tutorial, so the save states were moved underneath the Simulator in each example.

This fixes issue #1172.

### Details and comments
Using the current tutorial code, programming the following will lead to an (AttributeError: 'QuantumCircuit' object has no attribute 'save_state') error:

import numpy as np
from qiskit import QuantumCircuit
from qiskit import Aer, transpile
from qiskit.tools.visualization import plot_histogram, plot_state_city
import qiskit.quantum_info as qi

circ = QuantumCircuit(2)
circ.h(0)
circ.cx(0, 1)
circ.save_statevector()

simulator = Aer.get_backend('aer_simulator')
circ = transpile(circ, simulator)

result = simulator.run(circ).result()
statevector = result.get_statevector(circ)
plot_state_city(statevector, title='Bell state')


WiFisunset:
@SooluThomas @nonhermitian could you review this Pull Request?  This Pull Request also fixes issue #1172.

yaelbh:
`save_statevector` is an instruction of `QuantumCircuit` and should not be affected by the order of anything. You've changed the order of the lines
```
circ.save_statevector()
```
and
```
simulator = Aer.get_backend('aer_simulator')
```
These are independent lines and the code is equivalent. The second line doesn't do anything except for assigning a value to the `simulator` variable, which is not used in the first line.

A couple of additional comments:
1. With the usage of `save_statevector`, I think `get_statevector` needs to be replaced by something else.
2. As discussed in #1172, it appears that you have some mismatch with versions.

AshwinKul28:
Hi, @yaelbh I have added more explanation about it here. https://github.com/Qiskit/qiskit-tutorials/pull/1174 (although I need to close the PR for some another reason) but the details are here. I will add them here as well.

AshwinKul28:
### Summary
The circ.save_unitary() call is used to save the simulator state as a unitary matrix of the run circuit. In the example numbered as [10] in the simulators/1_aer_simulator notebook, the circ_save_unitary() call has been made before simulator initialization. Hence it is failing to run when a user tries to run exact same program.


### Details and comments
Since, the save_unitary() function call described as follows mentions that save_unitary() function saves the state of the simulator. (references: [ref1](https://qiskit.org/documentation/stubs/qiskit.circuit.QuantumCircuit.html#qiskit.circuit.QuantumCircuit.save_unitary), [ref2](https://qiskit.org/documentation/tutorials/simulators/1_aer_provider.html))

Screenshot 2021-04-28 at 1 11 27 AM

But according to the program if we don't call the simulator before the circ.save_unitary call, we cannot save the state of the simulator and hence QuantumCircuit is giving an error as follows:

Screenshot 2021-04-28 at 1 03 30 AM

Hence I have given the fix in the document as I'm calling save_unitary() function after the simulator initialisation.

Screenshot 2021-04-28 at 1 04 03 AM

yaelbh:
Ok, I can restore it in my environment. I don't think it's a desired behavior. Will check some more.

yaelbh:
https://github.com/Qiskit/qiskit-terra/issues/6346

WiFisunset:
I'm going to be updating this commit to issue a resolve found in https://github.com/Qiskit/qiskit-terra/issues/6346. 

And reverting the changes made previously where the lines were moved (to where the save states were under the simulator call).

In which calling 'from qiskit.providers.aer import Aer' is the solution to the save state problem.
qiskit poviders aer fix

WiFisunset:
@mtreinish could you review the changes I made to the docstrings regarding qiskit.providers.aer?

WiFisunset:
Bumping this PR. Could I get a review on the changes made?

tutorials/simulators/1_aer_provider.ipynb
```diff
@@ -30,8 +30,8 @@
     "import numpy as np\n",
     "\n",
     "# Import Qiskit\n",
-    "from qiskit import QuantumCircuit\n",
-    "from qiskit import Aer, transpile\n",
+    "from qiskit import QuantumCircuit, transpile\n",
+    "from qiskit.providers.aer import Aer\n",
```
mtreinish, line 34:
```suggestion
    "from qiskit_aer import Aer\n",
```

mtreinish (commented):
LGTM, just one small inline comment about the import using a legacy path

tutorials/simulators/1_aer_provider.ipynb
```diff
@@ -367,7 +367,8 @@
     "| `save_unitary` | Save the simulator state as unitary matrix of the run circuit | `\"automatic\"`, `\"unitary\"`|\n",
     "| `save_superop` | Save the simulator state as superoperator matrix of the run circuit | `\"automatic\"`, `\"superop\"`|\n",
     "\n",
-    "Note that these instructions are only supported by the Aer simulator and will result in an error if a circuit containing them is run on a non-simulator backend such as an IBM Quantum device."
+    "Note: These instructions are only supported by the Aer simulator and will result in an error if a circuit containing them is run on a non-simulator backend such as an IBM Quantum device."
+    "Note: The following must be imported to save the state of the simulator 'from qiskit.providers.aer import Aer'.
```
mtreinish, line 371: Actually besides the import path issue I mentioned before (which I missed here because of syntax highlighting) this also isn't actually 100% accurate. Any thing from `qiskit_aer` can be imported to add the monkey patched methods to `QuantumCircuit`.

Eric-Arellano:
Pardon that we didn't finish reviewing this. The Simulator tutorials now live in Aer and were removed from this repository in https://github.com/Qiskit/qiskit-tutorials/pull/1489

PR #1196, opened 2021-06-09
Closes #1189 Replacing 'state matrix' by 'density matrix'

### LINKED ISSUE #1189: use 'density matrix' instead of 'state matrix' in the plotting tutorial
the tutorials/circuits/2_plotting_data_in_qiskit.html tutorial uses 'state matrix' when talking about `plot_state_city` and `plot_state_hinton`. I think it's better to use 'density matrix' instead; the documentation of those methods also uses density matrix

shil-m:
### Summary
As described in issue #1189, "state matrix" has been replaced by "density matrix" in `plot_state_city` and 
 `plot_state_hinton` fotutorials/circuits/2_plotting_data_in_qiskit.html. Additionally I have also changed it at other plot functions.

### Details and comments
"State matrix" has a different physical meaning and doesn't go with the definition given in the documentation. Hence "density matrix" seems like a better option to use instead of "state matrix".


eliarbel (commented):
Looks good to me

eliarbel (approved):

Eric-Arellano (approved):
Pardon that we never merged this...thank you for the contribution, though!

PR #1200, opened 2021-06-13
Removed %matplotlib inline which was responsible for producing two outputs.

### LINKED ISSUE #1199: Two copies of each image in Transpiler Passes and Pass Manager documentation
**Problem -** Each images is printed twice after executing `draw('mpl')` , in the tutorial for Transpiler Passes and Pass Manager. 
**Possible Solution -** Look at the discussion in the issue number 1265 of Qiskit, the Author also faced a similar issue and there seems to be a possible solution there. 
Capture1

divshacker:
### Summary
fixes #1199 

### Details and comments
All details can be found in above linked issue.
### How removing `%matplotlib inline` fixes the problem. 
This discussion can be found here Qiskit/qiskit#1265


singhmeet11:
If possible we would like to extend this to solve issue #1201

SooluThomas (approved):
LGTM

Issue #1201, opened 2021-06-13
Two images due to %matplotlib inline in a few tutorials 


singhmeet11:
**Problem -** Each images is printed twice after executing draw('mpl') , in the tutorial for a few tutorials mainly - circuit basics, getting started with Qiskit, Device backend noise model simulations and Entanglement verification. \
This is in continuation with issue #1199.

**Possible solution -** Solution is to remove %matplotlib inline, this has been discussed in the issue [Qiskit/qiskit#1265](https://github.com/Qiskit/qiskit/issues/1265).

singhmeet11:
@divshacker why don't you take care of the other two since we came across this issue together. And should we remove `%matplotlib inline` from other tutorials too?? I actually went through all the tutorials and only 5 of them had this problem, some don't use `mpl` images some do but these were the only ones with repetitive images. Let me know what you think

singhmeet11:
And the last commit was by mistake, how can I remove that??

clasicus:
Hello! I don't use `%matplotlib inline` however I still have two circuits when I draw it by using `mpl` method. Moreover, when I use `plot_histogram` it also draws two histograms. 

qiskit               0.29.0
qiskit-aer           0.8.2
qiskit-aqua          0.9.4
qiskit-ibmq-provider 0.16.0
qiskit-ignis         0.6.0
qiskit-terra         0.18.1

Python 3.9.2

Thank you for your help!

divshacker:
Hey @clasicus You are using IBM quantum lab or your local system to run ?

clasicus:
> Hey @clasicus You are using IBM quantum lab or your local system to run ?

Good day! I use local jupyter-notebook. 
However, checked qiskit lab, for example, here: https://learn.qiskit.org/course/ch-labs/lab-1-quantum-circuits
And it works well.

divshacker:
Hey @clasicus It works fine for me, would you mind showing code and output which is showing this error

clasicus:
> Hey @clasicus It works fine for me, would you mind showing code and output which is showing this error

Of course, thank you.

For example:
`from qiskit import QuantumCircuit, Aer, execute` 
`from qiskit.visualization import plot_histogram`

`qc = QuantumCircuit(2,2)`

`qc.h(0)`
`qc.cnot(0,1)`
`qc.measure([0,1],[0,1])`

`qc.draw('mpl')`

`qc.draw('latex')`

[...]

PR #1206, opened 2021-06-28
Removing `%matplotlib inline` from different tutorials 

divshacker:
### Summary
fixes Qiskit/qiskit#1265
#1201 

### Details and comments
All details of this PR can be found in above linked issue


divshacker:
Done! all the `%matplotlib inline` removed.  @SooluThomas Please check and let me know

singhmeet11:
There were a few tutorials which were actually not using `.draw('mpl')` but had `%matplotlib inline`, @divshacker and I actually removed these ones too after having a discussion.\
**Reason** - As soon as someone would add matplot lib images, they would start to double up creating a sort of mess. This won't happen now.

divshacker:
> There were a few tutorials which were actually not using `.draw('mpl')` but had `%matplotlib inline`, @divshacker and I actually removed these ones too after having a discussion.
> **Reason** - As soon as someone would add matplot lib images, they would start to double up creating a sort of mess. This won't happen now.

Exactly! Also if one runs the `draw()` without `mpl` option or `%matplotlib inline` the figure produced by default is `mpl` only. (If we install the correct and ideal qiskit environment eg. IBM quantum lab which have all things preinstalled).

SooluThomas (approved):
LGTM

PR #1231, opened 2021-07-12
Fixed a code typo in 02_operators_overview.ipynb

elberttl:
### Summary
Fix a typo, in cell with  description "# Compose YX in front of the previous operator". The code compose XZ in front of the previous operator, while it should be YX. 

### Details and comments
`op.compose(XZ, qargs=[0, 2], front=True)` is changed with `op.compose(YX, qargs=[0, 2], front=True)`, and add the correct output.


Eric-Arellano (approved):
Pardon the late review. Good catch!

Eric-Arellano (approved):

PR #1237, opened 2021-07-13
Clearing up a repetition code (noise) tutorial example

clausia:
### Summary

In **noise/6_repetition_code.ipynb** tutorial.

According to this qiskit/qiskit#1025 issue explained by @Phionx, example 3 of the section "Running a repetition code" should change to `1 0 0000 0001 0000` instead of `0 1  0000 0001 0000` in order to be consistent, and thus correspond to a single qubit code that is flipped at the end of the line before the second round of syndrome measurements.

### Details and comments

Only the content of the related cell has been modified.

More details in https://github.com/Qiskit/qiskit/issues/1025


clausia:
In order to keep consistency with Qiskit notebook, now the example 3 is updated (it is actually one of the options proposed by @phionx)

HuangJunye:
This tutorial has been removed since you contributed. Sorry for the long delay in review.

PR #1241, opened 2021-07-19
adding compiler tools introduced in qv64 paper

ajavadia:


nonhermitian:
The tutorials do not make API calls like `IBMQ.load_account()`.  As I think you need that here, this tutorial should probably be hosted on the IQX until such time as the IBMQ provider is moved and hosts its own set of IBM Quantum specific tutorials.

tutorials/circuits_advanced/qv_tools.py
```diff
@@ -0,0 +1,146 @@
...
+        DynamicalDecoupling(instruction_durations, dd_sequence=[XGate(), XGate()])
+    ]
+
+    pm = PassManager()
+    pm.append(_map)                        # map to a subset of qubits, by choosing layout and inserting swaps
+    pm.append(_embed)                      # embed the mapped circuit onto the larger device coupling map
+    pm.append(_check_depth + _optimize,
+              do_while=_repeat_condition)  # translate to & optimize over hardware native gates
+    pm.append(_schedule)                   # dynamical decoupling on an as-late-as-possible schedule
+    return pm
```
mtreinish, line 63: Is there a reason to not put this in the tutorial text itself? The pass manager visualization is kind of hard to read and just having the code in the notebook is probably useful.

ajavadia: done

mtreinish:
One thing is to maybe avoid carrying around ~30MB of qpy data we can compress it with gzip and that cuts the size in about half for me locally. Then change the open lines to use `gzip.open()` from stdlib instead of the plain open. This will make the load slightly slower because it will decompress while reading, but should decrease the size of things people need to download and more importantly what we keep the git repo size down (as we keep around binary files in the repo forever and each change to the file will keep a new duplicate copy)

ajavadia:
closing this as it has been here for a while and might not be of interest anymore

PR #1243, opened 2021-07-31
replace execute with transpile + backend.run

ajavadia:
removed execute in places in favor of an explicit transpile step followed by run on backend.

didn't change the "noise" folder as qiskit-experiments will replace that entirely.

also changed a few places where `+` was still used rather than compose


tutorials/circuits_advanced/02_operators_overview.ipynb
```diff
@@ -519,7 +519,8 @@
    ],
    "source": [
     "backend = BasicAer.get_backend('qasm_simulator')\n",
-    "job = execute(circ, backend, basis_gates=['u1','u2','u3','cx'])\n",
+    "circ = transpile(circ, backend, basis_gates=['u1','u2','u3','cx'])\n",
+    "backend.run(circ)\n",
```
kdk, line 523:
```suggestion
    "job = backend.run(circ)\n",
```

kdk (approved):

PR #1244, opened 2021-08-05
[WIP] Add phase estimation tutorial

jlapeyre:
This adds a tutorial for phase estimation as implemented by subclasses of `PhaseEstimator`.

- [x] Basic example for each class
- [x] references (could possible add external reference, although this does not seem to be done in other tutorials)
- [ ] brief intro

### DIFF
--- a/tutorials/algorithms/10_phase_estimation.ipynb
+++ b/tutorials/algorithms/10_phase_estimation.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "id": "objective-pierce",
+   "id": "minute-compiler",
    "metadata": {},
    "source": [
     "# Phase Estimation"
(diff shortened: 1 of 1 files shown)


HuangJunye:
@jlapeyre It has been a while since the PR was opened. Do you still plan to add this tutorial?

jlapeyre:
Thanks, I forgot that this exists. If the result is valuable enough, and I'm not sure if it is, I'll finish it.

* This PR does not execute in its current state; due to bitrot I suppose. The first failure is at cell 16 after a couple of major features have been demonstrated correctly. I'm not sure, but I think this would not be hard to fix.

* This notebook uses opflow, which is scheduled to be replaced. I don't know how we expect the demand for the opflow API to evolve. The other tutorials in https://github.com/Qiskit/qiskit-tutorials/tree/master/tutorials/algorithms also use opflow, so it doesn't look like this phase estimation tutorial would be lagging behind.

I'll ask people who have been thinking about this.

jlapeyre:
I'll close this. As I said above may be worth salvaging some for a non-opflow version.

PR #1250, opened 2021-10-27
Add explanantion on Opflow's ``StateFn(H, is_measurement=True) @ state``

Cryoris:
### Summary

Add a paragraph explananing Opflow's ``StateFn(H, is_measurement=True) @ state`` notation.

### Details and comments

Also exchanges `StateFn(H).adjoint()` with `StateFn(H, is_measurement=True)` as  the latter doesn't take the complex conjugate of the observable and is thus safer to use.


HuangJunye:
@Cryoris It has been a while since this PR was submitted. Is this still relevant? Should I merge it?

Cryoris:
Since Opflow has been deprecated, we don't need this anymore 🙂

PR #1260, opened 2022-01-15
CI is failing because cvxpy 1.1 is not compatible with scs 3

1ucian0:
I think the failing in CI is because scs 3 is braking backwards compatibility and cvxpy 1.1 cannot handle it. For some reason, that also requires `scikit-learn`.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -38,6 +38,8 @@ steps:
     source /tmp/docs_build/bin/activate
     pip install -c constraints.txt -U qiskit jupyter sphinx nbsphinx sphinx_rtd_theme networkx
     pip install -c constraints.txt 'matplotlib<3.3'
+    # cvxpy 1.1 is not compatible with the last version of scs. Once updated, the scs dependecy can be removed
+    pip install -c constraints.txt -U 'scs<3' scikit-learn
     pip install -c constraints.txt -U qiskit[visualization] cvxpy
     pip install -c constraints.txt pyscf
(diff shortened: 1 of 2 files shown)


jakelishman (commented):
CVXPY 1.1.17+ is compatible with SCS 3 (https://github.com/cvxpy/cvxpy/releases/tag/v1.1.17), but in our `constraints.txt` file we're pinning `cvxpy<1.1.8`, because of issues with that release at the time requiring a large version of Numpy and old Qiskit components were using deprecated functionality (see #1127).  That's all fixed now, so we should be able to fix this by relaxing the constraint on `cvxpy`, rather than adding a new one on `scs`.

I suspect the trick with `scikit-learn` is that it's _always_ been required (it's opflow code that trips it), but previously some other dependency was installing it on our behalf.  We should add `scikit-learn` explicitly and probably remove the pin on `matplotlib` while we're at it (also in `environment.yml`) - #969 is long out-of-date, and Qiskit barely functions on MPL 3.2 any more.

mtreinish:
> I suspect the trick with `scikit-learn` is that it's _always_ been required (it's opflow code that trips it), but previously some other dependency was installing it on our behalf.

It used to be installed by `qiskit-aqua` when we pip installed `qiskit` which had scikit-learn in it's requirements list.

jakelishman (approved):

1ucian0:
I completely missed the `constraints.txt` entry. Thanks for pointing it out.

SooluThomas (approved):

PR #1265, opened 2022-02-14
`Parameter`s to `Parameters`

NG-Glen:
`Parameter`s to `Parameters`

### DIFF
--- a/tutorials/circuits_advanced/01_advanced_circuits.ipynb
+++ b/tutorials/circuits_advanced/01_advanced_circuits.ipynb
@@ -303,7 +303,7 @@
    "metadata": {},
    "source": [
     "All circuit parameters must be bound before sending the circuit to a backend. This can be done as follows:\n",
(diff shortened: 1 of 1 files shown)


SooluThomas:
I think `Parameter`s is the right way because it is taking about the type `Parameter`

NG-Glen:
<p>Please reference https://sooluthomas.github.io/testTranslation/terra/custom_gates.html</p><p>Binding Parameters to Values¶</p><p> </p><blockquote>The <code>bind_parameters</code> method accepts a dictionary mapping <code>Parameters</code> to values, and returns a new circuit with each parameter replaced by its corresponding value.</blockquote><p> </p><blockquote>Partial binding is supported, in which case the returned circuit will be parameterized by any <code>Parameters</code> which were not mapped to a value.</blockquote>

NG-Glen:
<p>It's very clear now, so many <code>Parameter</code>s is the right way, but it's very confusing, could you change to "<code>Parameter</code> objects"</p><p> </p>

PR #1271, opened 2022-02-25
2 change to {2}

NG-Glen:
2 change to {2}


SooluThomas:
Why is this change necessary when the docs is originally rendered correctly? 

image

PR #1284, opened 2022-03-17
Fix small render error

clausia:
### Summary

Sphinx doesn't render a code block correctly because it is inside an italicized string.

### Details and comments

* Placing a block of code inside text in italics causes Sphinx to render the block of code incorrectly, as seen in the image below (Figure 1).
The way to fix this is to break the text into italics so that it doesn't include the block of code.
This is presented in the [GPU Simulation](https://qiskit.org/documentation/tutorials/simulators/1_aer_provider.html#GPU-Simulation) section.

* A missing comma is added (Figure 2).
This is in the [Simulation Precision](https://qiskit.org/documentation/tutorials/simulators/1_aer_provider.html#Simulation-Precision)

-------
Figure 1
image

Figure 2
image


1ucian0 (approved):
Thanks!

PR #1297, opened 2022-03-31
Update: Plot histogram to Plot Histogram

NG-Glen:


HuangJunye (approved):
Thanks for the quick fix!

PR #1301, opened 2022-04-07
qubit {j} to qubit $Q_{j}$ more readable

NG-Glen:


HuangJunye (approved):
Looks good to me. Thanks for your contribution!

PR #1306, opened 2022-04-30
Fixed typo for identity gate

ChihHanHuang:
Fixes typo of "The identity gate is $Id = u0(1)$."
It should be "The identity gate is $Id = u1(0)$."

### Summary

Fixed typo for identity gate

### Details and comments

There is a typo of "The identity gate is $Id = u0(1)$."
It should be "The identity gate is $Id = u1(0)$."

ref: https://qiskit.org/documentation/stubs/qiskit.circuit.library.U1Gate.html


HuangJunye (approved):
Thanks for spotting this and making the contribution!

Issue #1310, opened 2022-05-04
Endianness in text of Circuits tutorial 3


aeddins-ibm:
![image](https://user-images.githubusercontent.com/60495383/166814323-2308e237-b546-4263-bf56-6455e75e581d.png)

I think the parenthetical at the end of the highlighted sentence above should be changed to "little-endian". 

I often mix this up. But, following the sentence, the least significant bit (the little bit) is on the right (at the end).

Also, other online info/docs state that Qiskit uses little-endian:
https://quantumcomputing.stackexchange.com/questions/8244/big-endian-vs-little-endian-in-qiskit
https://qiskit.org/documentation/stubs/qiskit.circuit.library.CXGate.html

InfamousPlatypus:
if aeddins doesn't want to deal with this I would love to start getting involved in something.

HuangJunye:
@aeddins-ibm Would you like to make the PR yourself or would you like @InfamousPlatypus to do that? Please let us know!

aeddins-ibm:
Please go ahead and make the PR @InfamousPlatypus . Thanks!

SahilDahat:
Is this issue resolved? The issue arose due to a discrepancy between the little-endian representation and the big-endian notation specified in the documentation, correct?

aeddins-ibm:
LGTM. Belated thanks to @InfamousPlatypus !

![image](https://github.com/Qiskit/qiskit-tutorials/assets/60495383/4f15be77-a307-42c5-af3c-487779f8bbfd)

PR #1319, opened 2022-05-18
#1310 Incorrect Endianness Fix

InfamousPlatypus:
### Summary
Fixed issue #1310


InfamousPlatypus:
> @InfamousPlatypus Thanks for the fix. Can you please also do a quick search to see if there is other mentions of endianness in this and other tutorials and make them consistent? That would be really helpful. Thank you!


That was all.

HuangJunye (approved):
LGTM. Thanks for your contributions

Issue #1315, opened 2022-05-18
Remove index page


HuangJunye:
### Informations

- **Qiskit version**:
- **Python version**:
- **Operating system**:

### What is the current behavior?

Each section in the tutorial page has its own index page showing up as an icon similar to the tutorials itself which is confusing.

### Steps to reproduce the problem

Go to https://qiskit.org/documentation/tutorials.html#quantum-circuits, click the last icon "circuit tutorials".

<img width="843" alt="image" src="https://user-images.githubusercontent.com/7631333/169010058-5de91fd2-3899-4e10-9888-7700a671d8f5.png">

It will lead to this page https://qiskit.org/documentation/tutorials/circuits/index.html which contain the same tutorials in the section but in its own page.

### What is the expected behavior?
Remove index page of each section in https://qiskit.org/documentation/tutorials.html

### Suggested solutions
Either remove index.rst in each folders of qiskit-tutorials repo or update the glob rule in [tutorials.rst](https://github.com/Qiskit/qiskit/blob/master/docs/tutorials.rst) in qiskit meta package

mtreinish:
The index pages are required for building the tutorials in standalone docs builds. Without the index files for each section the tutorials will not be included in the toctree for the build. This will cause warnings which will be fatal in downstream builds (such as in qiskit component CI) and a docs build output where you can't navigate to the tutorials (assuming it builds them).

HuangJunye:
I suspected that. Is it possible to exclude index.rst in the glob rules under each section in qiskit meta package repo?

HuangJunye:
This PR https://github.com/Qiskit/qiskit-tutorials/pull/1234 tried to address this issue but it was put on hold. Let me investigate a bit further.

PR #1322, opened 2022-05-19
Fix too many `index.rst` problem (#1315)

SaumyaBhushan:
Closes https://github.com/Qiskit/qiskit-tutorials/issues/1315. As explained there, these `index.rst` pages result in our Tutorials home page having extra entries that simply link to that section's other entries.

This PR fixes it by mirroring `tutorials.rst` from `qiskit-metapackage`. Our `index.html` now directly calls `nbgallery` with `glob`, rather than referring to `index.rst` using normal Sphinx mechanisms. That results in this tutorials repo rendering like this:

Screenshot 2023-07-11 at 9 54 04 AM

The metapackage ignores this repo's `index.rst` and only copies over the `tutorials/` folder. So, it will copy over the same content as before, only now without the `index.rst` files in each subfolder.


HuangJunye:
Thanks for your contributions. I need to do a local build to see if removing the index pages is indeed the right approach. The qiskit documentation building is a bit complicated. I think the index page might still be useful when the tutorials are built on its own. An alternative approach is to ignore the index page in the qiskit meta package.

SaumyaBhushan:
Sure @HuangJunye

HuangJunye:
Closing this PR because qiskit-tutorials content is going to be migrated to qiskit-terra: https://github.com/Qiskit/qiskit-tutorials/issues/1473 Thanks a lot for your interest in contributing all the same!

Eric-Arellano:
> Closing this PR because qiskit-tutorials content is going to be migrated to qiskit-terra: https://github.com/Qiskit/qiskit-tutorials/issues/1473 Thanks a lot for your interest in contributing all the same!

Actually, no, let's keep this open! I was today going to try making this same page because this is an annoying issue in the metapackage and fixing this will help in the metapackage migration. I didn't realize this was already open!

Thank you @SaumyaBhushan and pardon how long it took for a review. I will try this out locally today and merge once confirmed it works how we want.

javabster (approved):
LGTM 🚀

PR #1328, opened 2022-05-26
Fix for Ambiguity in Issue #1261

### LINKED ISSUE #1261: tutorials/textbook/01_IQPE.ipynb documentation error
In the section "Conditined gates: the c_if method":
1) the section header "Conditined" is not a word. did you mean "Conditioned" or "Conditional"?
2) the paragraph that ends "when the results of the measurement of 𝑞0 and 𝑞1 are both 1." is incorrect? Shouldn't it say ""when the results of the measurement of 𝑞1 and 𝑞2 are both 1." The rest of the section is consistent with this change.
3) further down, at "since it does not matter what is in the rightmost position.". Shouldn't it say "...in the leftmost position"?

It seems the example originally had bits q0 and q1 as the conditionals, and then they were changed to q1 and q2 but not all text was changed accordingly, so there may be other typos in addition to these.

padraignix:
### Summary

Wording change on IQPE notebooks to replace potentially confusing statement with concrete verbiage.

### Details and comments

Fixes #1261 - Issue should be closable post introduction of this PR

### DIFF
--- a/tutorials/algorithms/09_IQPE.ipynb
+++ b/tutorials/algorithms/09_IQPE.ipynb
@@ -116,7 +116,7 @@
    "source": [
     "We want to apply the $X$ gate, only if both the results of the measurement of $q_0$ and $q_1$ are $1$. We can do this using the c_if method, conditioning the application of $X$ depending on the value passed as argument to c_if.\n",
     "\n",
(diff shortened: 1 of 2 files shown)


HuangJunye (approved):
Thanks for working on this issue!

PR #1356, opened 2022-09-12
Update pulse gate tutorial for changes in fake backends

mtreinish:
### Summary

In Qiskit/qiskit-terra#8665 very large pulse properties for long retired 20q backends are removed to save disk space and reduce overhead. However the pulse gates tutorial was still relying on these fake backends having pulse properties. This commit updates the tutorial to use a different backend to unblock that PR. Additionally, the deprecate import statements used in the tutorial are updated to avoid the deprecation warning.

### Details and comments


1ucian0 (approved):
Thanks!

PR #1357, opened 2022-09-16
qiskit.providers.aer is qiskit_aer in Aer 0.11

1ucian0:
In the `qiskit-aer` most recent release 0.11, the import path is changing from `qiskit.providers.aer` is `qiskit_aer`. This is a silent change right now, but it will turn into a warning with `qiskit-terra 0.22` (planned for Oct 11th, 2022)

This PR change all the import paths and the requirement.


requirements-dev.txt
```diff
@@ -1,4 +1,5 @@
 qiskit
+qiskit-aer>=0.11
```
mtreinish, line 2: While aer is included in the metapackage you shouldn't do this, it's possible to cause a conflict with `qiskit` as `qiskit` pins aer to a fixed version. I would change `qiskit` to `qiskit>=0.38.0` for the time being. In a future date where aer is not in the metapackage we'll need to do this.

mtreinish (dismissed):
Besides the inline comment there are also some tutorials that are using `qiskit.Aer` that you'll want to update to `qiskit_aer.Aer` too.

requirements-dev.txt
1ucian0: done in 7dc10cdc5bdf5bf47601ba54fbb004eebcbc2a5f

HuangJunye (approved):

PR #1375, opened 2022-11-15
Bell state has coefficient 1/sqrt(2) in front

brian36:
Notebook has an error - it shows bell state with coefficient 1/2 in front, but it's 1/sqrt(2)

### Summary
This tutorial notebook has a typo - it says the bell state has coefficient 1/2 in front.  Instead it should be 1/sqrt(2).  

### Details and comments
I made this small change in the notebook.


HuangJunye (approved):
Thanks for your contributions!

Issue #1363, opened 2022-10-27
Update 04_vqe_advanced.ipynb


dlasecki:
### What is the expected behavior?

Resolve the following problems in https://github.com/Qiskit/qiskit-tutorials/blob/master/tutorials/algorithms/04_vqe_advanced.ipynb:

- Uses quantum instance, opflow in examples,
- Uses old VQE, gradient framework,
- Discusses outdated custom expectation.

ElePT:
I can take this issue.

ElePT:
This issue has been closed by https://github.com/Qiskit/qiskit-tutorials/pull/1376

HuangJunye:
Thanks for keeping track of this, Elena!

Issue #1362, opened 2022-10-27
Update 03_vqe_simulation_with_noise.ipynb


dlasecki:
### What is the expected behavior?

Resolve the following problems in https://github.com/Qiskit/qiskit-tutorials/blob/master/tutorials/algorithms/03_vqe_simulation_with_noise.ipynb:

- Uses quantum instance, opflow in examples,
- Uses from qiskit.providers.aer.noise import NoiseModel; might switch it to the BackendEstimator and BackendSampler from Aer,
- Uses from qiskit.utils.mitigation import CompleteMeasFitter; might use readout mitigation from primitives,
- Uses old VQE.

ElePT:
I can take this issue.

ElePT:
This issue has been closed by https://github.com/Qiskit/qiskit-tutorials/pull/1376

HuangJunye:
Thanks!

PR #1376, opened 2022-11-16
Update VQE tutorials 02-04

### LINKED ISSUE #1361: Update 02_vqe_convergence.ipynb
### What is the expected behavior?

Resolve the following problems in https://github.com/Qiskit/qiskit-tutorials/blob/master/tutorials/algorithms/02_vqe_convergence.ipynb:

- Uses quantum instance, opflow in examples.
- Uses old VQE, gradients.

ElePT:
### Summary
This PR updates the VQE tutorials to the new algorithm syntax and removes outdated explanations based on deprecated functionality (`Expectation`, `QuantumInstance`...). This change allowed me to merge tutorials 02 and 04 into a new "VQE Advanced Options" tutorial. 

Closes #1361, #1362, #1363

### Details and comments
Merging tutorials `02` and `04` leaves a space for a new `04` tutorial that I think should cover the `VQD` algorithm.


tutorials/algorithms/02_vqe_advanced_options.ipynb
```diff
@@ -0,0 +1,572 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import logging\n",
+    "\n",
+    "# uncomment to set up logging\n",
+    "# logging.basicConfig(level=logging.INFO)\n",
+    "# logging.getLogger(\"qiskit.algorithms.minimum_eigen_solvers.vqe\").setLevel(logging.INFO)"
+   ]
+  },
```
woodsp-ibm, line 353: In the algorithms in Aqua we used to add more extensive logging than seems to get added nowadays. The example shown before would work and the logging below was a couple of lines copy/pasted from the output. As it stands the logging commented out above there is still enabling it for the old vqe not the new. BUT even if it were enabled the output as shown below (which is still from old VQE) would no longer occur as  I do not see there is logging in the energy evaluation function anymore. While there is logging there the only output now is minimal - in looking it logs when trying to resize an ansatz and at the end when the optimization is complete - so just 2 places. As such maybe we should just drop this entire logging section.

randasbeity:
Thank you.

tutorials/algorithms/02_vqe_advanced_options.ipynb
```diff
@@ -0,0 +1,572 @@
...
+    "INFO:qiskit.algorithms.minimum_eigen_solvers.vqe:Energy evaluation returned [-1.75399268] - 6.61016 (ms), eval count: 4"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Initial point\n",
+    "\n",
+    "The `initial_point`option allows the optimization to begin at a given point, defined by a list of parameter values that will configure the ansatz. By default, the initial point is `None`, which means that VQE will choose it based on the initial state of the ansatz. If the ansatz does not provide a preferred state, a random initial point that fits within the ansatz bounds will be chosen. If an initial point is supplied by the user, it will take priority and be used - note though it must match in length to the number of parameters in the ansatz circuit.\n",
```
woodsp-ibm, line 372: In terms of the initial point I think nowadays its either a randomly chosen point or explicitly chosen/input by a user.

We used to have the possibility that an anstaz could indicate a preferred initial point, if it were given an initial state. For instance HF + UCC would have done this in the past. This is no longer the case. Yes it still will look to the ansatz in case it has any parameter bounds defined, but that is just to ensure that the random point will be within these bounds and hence valid. I think this is more than a user needs to know though and just saying its random by default should suffice.

tutorials/algorithms/02_vqe_advanced_options.ipynb
```diff
@@ -0,0 +1,572 @@
...
+     "name": "#%%\n"
+    }
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "{   'aux_operators_evaluated': None,\n",
+      "    'cost_function_evals': 9,\n",
```
woodsp-ibm, line 441: In the old tutorial the runs were more equivalent. If you look the one took 65 evals with SLSQP and the one at the optimal point just 9. Both here seem to take just 9. I think thats because before the internal optimizers finite diff was used which computes the gradient with the objective function - which is the latter case here too when the optimal point is given. Which explains the difference in time despite the number or evals in this case being the same. Maybe it would be better to run this latter one using the gradient to be more directly comparable with the reference result. I imagine then it would be just be 1 eval since its at the optimal point the gradient will cause it to to terminate since its already at the minumum.

tutorials/algorithms/03_vqe_simulation_with_noise.ipynb
```diff
@@ -186,158 +229,117 @@
...
-    "print(f'VQE on Aer qasm simulator (with noise): {result1.eigenvalue.real:.5f}')\n",
-    "print(f'Delta from reference energy value is {(result1.eigenvalue.real - ref_value):.5f}')"
+    "print(noise_model)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Once the noise model is defined, we can train our VQE using a noisy Aer Estimator. Please note that this simulation will take longer than the noiseless one."
```
woodsp-ibm, line 274: I don't think I would use the term `train` - we just run VQE right in the presence of noise

Maybe: Once the noise model is defined, we can run VQE using an Aer Estimator where we pass the noise model to the underlying simulator using the backend_options dictionary.

tutorials/algorithms/03_vqe_simulation_with_noise.ipynb
```diff
@@ -186,158 +229,117 @@
...
-    "    algorithm_globals.random_seed = seed\n",
-    "    qi = QuantumInstance(backend=backend, seed_simulator=seed, seed_transpiler=seed,\n",
-    "                         coupling_map=coupling_map, noise_model=noise_model, \n",
-    "                         measurement_error_mitigation_cls=CompleteMeasFitter, \n",
-    "                         cals_matrix_refresh_period=30)\n",
-    "\n",
-    "    def store_intermediate_result2(eval_count, parameters, mean, std):\n",
-    "        counts2.append(eval_count)\n",
-    "        values2.append(mean)\n",
+    "vqe = VQE(noisy_estimator, ansatz, optimizer=spsa, callback=store_intermediate_result)\n",
```
woodsp-ibm, line 326: Now we could just re-use the prior VQE instance right, setting the estimator and possibly the new callback if we want to retain the former results (rather that say just clearing its lists and reusing that same callback). Might be more interesting as a different way to do things rather than seeing VQE's built out from scratch each time?

tutorials/algorithms/03_vqe_simulation_with_noise.ipynb
```diff
@@ -349,57 +351,66 @@
...
+   "source": [
+    "## Summary\n"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Lets bring the results together here for a summary. \n",
+    "We produced a reference value using a classical algorithm and then proceeded to run VQE on an Aer simulator. While the simulation is ideal (no noise) there is so called shot-noise due to sampling - increasing the number of shots reduces this as more and more samples are taken, but shots were left at the default of 1024 and we see a small effect in the outcome. \n",
```
woodsp-ibm, line 373: Well for the Estimator shots was left at the default of None. Does this internally default to 1024 - that is certainly what the defaults when you used shots in the past to run circuits directly. I did not see anything in the Aer Estimator docs as to what to expect - maybe it drops through to that same default.

tutorials/algorithms/03_vqe_simulation_with_noise.ipynb
```diff
@@ -349,57 +351,66 @@
...
     "\n",
-    "We produced a reference value using a classical algorithm and then proceeded to run VQE on a qasm simulator. While the simulation is ideal (no noise) there is so called shot-noise due to sampling - increasing the number of shots reduces this as more and more samples are taken, but shots was left at the default of 1024 and we see a small effect in the outcome. \n",
+    "Then, we added noise using a model taken off a real device and can see the result is affected. \n",
     "\n",
-    "Then we added noise using a model taken off a real device and can see the result is affected. Finally we added measurement noise mitigation which adjusts the results in an attempt to alleviate the affect of noise in the classical equipment measuring the qubits."
+    "In the future, we could apply error mitigation techniques to alleviate the effect of noise in simulation."
```
woodsp-ibm, line 377: This future statement only relates to possibility of adding mitigation support to Are primitives right. Maybe we should note that if you are using VQE with the runtime then the Estimator there offers noise mitigation via the resilience_level and to see https://qiskit.org/documentation/partners/qiskit_ibm_runtime/ for more details?

tutorials/algorithms/02_vqe_advanced_options.ipynb
ElePT: Yes, that's true, I did not even try out the logging. Removing.

tutorials/algorithms/02_vqe_advanced_options.ipynb
ElePT: This makes sense, I was a bit confused about what "preferred by the ansatz" meant. I see why now.

tutorials/algorithms/02_vqe_advanced_options.ipynb
ElePT: Good point, I just assumed the new gradients modified the number of evals.

tutorials/algorithms/03_vqe_simulation_with_noise.ipynb
ElePT: Yes, this was from the old tutorial, I was still a bit confused by it but assumed it was right (learning I shouldn't assume anymore 😅)

tutorials/algorithms/03_vqe_simulation_with_noise.ipynb
ElePT: Good idea.

tutorials/algorithms/03_vqe_simulation_with_noise.ipynb
ElePT: Mmm true, I did not look very deep into the Aer Estimator default, I will test it real quick.

tutorials/algorithms/02_vqe_advanced_options.ipynb
```diff
@@ -0,0 +1,541 @@
...
+    "from qiskit.primitives import Estimator\n",
+    "\n",
+    "estimator = Estimator()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We will now see the VQE callback usage over a set of optimizers for comparison. The minimum energy of the H2 Hamiltonian can be found quite easily, so the value of maxiters does not have to be very large. We will once again use `TwoLocal` as our selected trial wavefunction (i.e. ansatz)."
```
dlasecki, line 84:
```suggestion
    "We will now see the VQE callback usage over a set of optimizers for comparison. The minimum energy of the H2 Hamiltonian can be found quite easily, so the value of `maxiters` does not have to be very large. We will once again use `TwoLocal` as our selected trial wavefunction (i.e. ansatz)."
```

tutorials/algorithms/02_vqe_advanced_options.ipynb
```diff
@@ -0,0 +1,541 @@
...
+    "from qiskit.utils import algorithm_globals\n",
+    "\n",
+    "# we will iterate over these different optimizers\n",
+    "optimizers = [COBYLA(maxiter=80), L_BFGS_B(maxiter=60), SLSQP(maxiter=60)]\n",
+    "converge_counts = np.empty([len(optimizers)], dtype=object)\n",
+    "converge_vals = np.empty([len(optimizers)], dtype=object)\n",
+    "\n",
+    "for i, optimizer in enumerate(optimizers):\n",
+    "    print(\"\\rOptimizer: {}        \".format(type(optimizer).__name__), end=\"\")\n",
+    "    algorithm_globals.random_seed = 50\n",
```
dlasecki, line 116: Do we need to set it in every iteration?

tutorials/algorithms/02_vqe_advanced_options.ipynb
```diff
@@ -0,0 +1,541 @@
...
+    "    converge_vals[i] = np.asarray(values)\n",
+    "\n",
+    "print(\"\\rOptimization complete      \");"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Now from the callback data we stored we can plot the energy value at each objective function call each optimizer makes. An optimizer using a finite difference method for computing gradient has that characteristic step like plot where for a number of evaluations it is computing the value for close by points to establish a gradient (the close by points having very similar values whose difference cannot be seen on the scale of the graph here)."
```
dlasecki, line 138:
```suggestion
    "Now from the callback data we stored we can plot the energy value at each objective function call each optimizer makes. An optimizer using a finite difference method for computing gradient has that characteristic step-like plot where for a number of evaluations it is computing the value for close by points to establish a gradient (the close by points having very similar values whose difference cannot be seen on the scale of the graph here)."
```

tutorials/algorithms/02_vqe_advanced_options.ipynb
```diff
@@ -0,0 +1,541 @@
...
+    "pylab.ylabel(\"Energy\")\n",
+    "pylab.title(\"Energy convergence for various optimizers\")\n",
+    "pylab.legend(loc=\"upper right\");"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Finally since the above problem is still easily tractable classically we can use `NumPyMinimumEigensolver` to compute a reference value for the solution. We can now plot the difference from the resultant exact solution as the energy converges with `VQE` towards the minimum value which should be that exact classical solution."
```
dlasecki, line 180:
```suggestion
    "Finally, since the above problem is still easily tractable classically we can use `NumPyMinimumEigensolver` to compute a reference value for the solution. We can now plot the difference from the resultant exact solution as the energy converges with `VQE` towards the minimum value which should be that exact classical solution."
```

tutorials/algorithms/03_vqe_simulation_with_noise.ipynb
```diff
@@ -4,44 +4,23 @@
...
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Noisy simulation will be demonstrated here with VQE, finding the minimum (ground state) energy of an Hamiltonian, but the technique applies to any quantum algorithm from Qiskit.\n",
-    "\n",
-    "So for VQE we need a qubit operator as input. Here, once again, we will take a set of paulis that were originally computed by Qiskit Nature, for an H2 molecule, so we can quickly create an Operator."
+    "The algorithm of choice is once again VQE, where the task consists on finding the minimum (ground state) energy of a Hamiltonian. As shown in previous tutorials, VQE takes in a qubit operator as input. Here, we will take a set of paulis that were originally computed by Qiskit Nature for the H2 molecule, using the `SparsePauliOp` class."
```
dlasecki, line 18:
```suggestion
    "The algorithm of choice is once again VQE, where the task consists on finding the minimum (ground state) energy of a Hamiltonian. As shown in previous tutorials, VQE takes in a qubit operator as input. Here, we will take a set of Pauli operators that were originally computed by Qiskit Nature for the H2 molecule, using the `SparsePauliOp` class."
```

tutorials/algorithms/03_vqe_simulation_with_noise.ipynb
```diff
@@ -53,25 +32,31 @@
...
-    "print(f'Number of qubits: {H2_op.num_qubits}')"
+    "print(f\"Number of qubits: {H2_op.num_qubits}\")"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "As the above problem is still easily tractable classically we can use NumPyMinimumEigensolver to compute a reference value so we can compare later the results. "
+    "As the above problem is still easily tractable classically we can use `NumPyMinimumEigensolver` to compute a reference value so we can compare later the results. "
```
dlasecki, line 54:
```suggestion
    "As the above problem is still easily tractable classically we can use `NumPyMinimumEigensolver` to compute a reference value so we can compare the results later. "
```

tutorials/algorithms/03_vqe_simulation_with_noise.ipynb
```diff
@@ -186,158 +228,117 @@
...
-    "\n",
-    "Note: simulation with noise will take longer than without noise.\n",
+    "Now, we will add noise to our simulation. In particular, we will extract a noise model from a (fake) device. As stated in the introduction, it is also possible to create custom noise models from scratch, but this task is beyond the scope of this notebook.\n",
     "\n",
-    "Terra Mock Backends:\n",
-    "\n",
-    "We will use real noise data for an IBM Quantum device using the date stored in Qiskit Terra. Specifically, in this tutorial, the device is ibmq_vigo."
+    "First we need to get an actual device backend and from its `configuration` and `properties` we can setup a coupling map and a noise model to match the device. Note: We can also use this coupling map as the entanglement map for the variational form if we choose."
```
dlasecki, line 233:
```suggestion
    "First, we need to get an actual device backend and from its `configuration` and `properties` we can setup a coupling map and a noise model to match the device. Note: We can also use this coupling map as the entanglement map for the variational form if we choose to."
```

tutorials/algorithms/03_vqe_simulation_with_noise.ipynb
```diff
@@ -349,57 +350,71 @@
...
   },
   {
-   "cell_type": "code",
-   "execution_count": 11,
+   "cell_type": "markdown",
    "metadata": {},
+   "source": [
+    "We can notice that, while the noiseless simulation's result is closer to the exact reference value, there is still some difference. This is due to the sampling noise, introduced by limiting the number of shots to 1024. A larger number of shots would decrease this sampling error and close the gap between these two values. \n",
+    "\n",
+    "As for the noise introduced by real devices (or simualted noise models), it could be tackled through a wide variety of error mitigation techniques. The [Qiskit Runtime Primitives](https://qiskit.org/documentation/partners/qiskit_ibm_runtime/) have enabled error mitigation through the `resilience_level` option. This option is currently available for remote simulators and real backends accessed via the Runtime Primitives, you can consult [this tutorial](https://qiskit.org/documentation/partners/qiskit_ibm_runtime/tutorials/Error-Suppression-and-Error-Mitigation.html) for further information."
```
dlasecki, line 402:
```suggestion
    "As for the noise introduced by real devices (or simulated noise models), it could be tackled through a wide variety of error mitigation techniques. The [Qiskit Runtime Primitives](https://qiskit.org/documentation/partners/qiskit_ibm_runtime/) have enabled error mitigation through the `resilience_level` option. This option is currently available for remote simulators and real backends accessed via the Runtime Primitives, you can consult [this tutorial](https://qiskit.org/documentation/partners/qiskit_ibm_runtime/tutorials/Error-Suppression-and-Error-Mitigation.html) for further information."
```

tutorials/algorithms/02_vqe_advanced_options.ipynb
woodsp-ibm: Yes, so that the starting point and the conditions for each run are identical when random - we are just looking for what happens with the different optimizers in each loop compared to one another.

tutorials/algorithms/02_vqe_advanced_options.ipynb
woodsp-ibm: The variable is `maxiter` so if we want to have it highlighted I guess it should be correct

ElePT:
```suggestion
    "We will now see the VQE callback usage over a set of optimizers for comparison. The minimum energy of the H2 Hamiltonian can be found quite easily, so the number of iterations (`maxiter`) does not have to be very large. We will once again use `TwoLocal` as our selected trial wavefunction (i.e. ansatz)."
```

ElePT: Fixed

woodsp-ibm (approved):

HuangJunye:
@Guillermo-Mijares-Vilarino Can you please review this PR?

Guillermo-Mijares-Vilarino:
The content LGTM. The tutorials look cleaner and some of the explanations are better now.

I only have a small general comment. In general we are trying to encourage the use of "you" instead of "we" for the tutorials, to focus more on the agency of the user instead of being taken by the hand by us.

ElePT:
Thanks for the feedback! I changed most of the we's to you's :)

SooluThomas (approved):
LGTM

Issue #1373, opened 2022-11-08
Add pVQD tutorial.


dlasecki:
### What is the expected behavior?

The goal of this issue is to introduce a new tutorial that covers a pVQD algorithm from Qiskit Terra algorithms package. The tutorial should demonstrate how to use the algorithm with primitives.

PR #1382, opened 2022-12-09
Add ``PVQD`` tutorial

Cryoris:
### Summary

Add a tutorial on the p-VQD algorithm implemented in Qiskit as ``PVQD``.

Closes: #1373


tutorials/algorithms/10_pvqd.ipynb
```diff
@@ -0,0 +1,591 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The example we're looking at is the time evolution of the $|00\\rangle$ state under the Hamiltonian\n",
+    "\n",
+    "$$\n",
+    "    H = 0.1 Z_1 Z_2 + X_1 + X_2,\n",
+    "$$\n",
+    "\n",
+    "which is an Ising Hamiltonian on two neighboring spins, up to a time $T=1$ and we want to keep track of the total magnetization $M = Z_1 + Z_2$."
```
ElePT, line 32:
```suggestion
    "which is an Ising Hamiltonian on two neighboring spins, up to a time $T=1$, where we want to keep track of the total magnetization $M = Z_1 + Z_2$."
```

tutorials/algorithms/10_pvqd.ipynb
```diff
@@ -0,0 +1,591 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from qiskit.quantum_info import SparsePauliOp\n",
+    "\n",
+    "final_time = 1\n",
+    "hamiltonian = SparsePauliOp.from_sparse_list([\n",
+    "    (\"ZZ\", [0, 1], 0.1), (\"X\", [0], 1), (\"X\", [1], 1),\n",
+    "], num_qubits=2)\n",
+    "observable = SparsePauliOp([\"ZI\", \"IZ\"])"
```
ElePT, line 47: I think it's more readable if you explicitly include `magnetization` in the variable name, maybe `magnetization_observable`? (I know it's a bit long, open to other ideas)

tutorials/algorithms/10_pvqd.ipynb
```diff
@@ -0,0 +1,591 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Next, we need to choose the parameterized ansatz we project the update steps onto. We have different choices here, but for real time evolution an ansatz that contains building blocks of the evolved Hamiltonian usually performs very well."
```
ElePT, line 54:
```suggestion
    "After defining our hamiltonian and observable, we need to choose the parameterized ansatz we project the update steps onto. We have different choices here, but for real time evolution, an ansatz that contains building blocks of the evolved Hamiltonian usually performs very well."
```

tutorials/algorithms/10_pvqd.ipynb
```diff
@@ -0,0 +1,591 @@
...
+   "outputs": [],
+   "source": [
+    "pvqd.optimizer = gd"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "**Warning:** The following cell will run a few minutes."
```
ElePT, line 421:
```suggestion
    "**Warning:** The following cell will take a few minutes to run."
```

tutorials/algorithms/10_pvqd.ipynb
```diff
@@ -0,0 +1,591 @@
...
+    "plt.xlabel(\"time $t$\")\n",
+    "plt.ylabel(\"energy $E$\")\n",
+    "plt.title(\"Energy over time\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We can observe here, that the energy does vary quite a lot! But as we mentioned before, p-VQD does not preserve the energy."
```
ElePT, line 476:
```suggestion
    "We can observe here that the energy does vary, quite a lot! But as we mentioned before, p-VQD does not preserve the energy."
```

ElePT (commented):
I added a few suggestions but overall the tutorial looks great! Maybe to get an even better result, you could add a final paragraph that acts as a conclusion, because I get the feeling that the tutorial ends a bit suddenly.

tutorials/algorithms/10_pvqd.ipynb
```diff
@@ -0,0 +1,591 @@
+{
```
Guillermo-Mijares-Vilarino, line 1: <p>Should we expect users to be familiar with Trotterization? I'd include at least a link to a page that explains it.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Qiskit/qiskit-tutorials/pull/1382/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

Guillermo-Mijares-Vilarino, line 1: <p>I think it would be nice to mention explicitly that you'll keep track of both the Hamiltonian and the magnetization (with the <code>aux_operators</code> argument of the <code>TimeEvolutionProblem</code>).</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Qiskit/qiskit-tutorials/pull/1382/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

Guillermo-Mijares-Vilarino, line 1: <p>Here I'd avoid to include any reference to the ansatz, as it doesn't have much to do with the exact calculation (it's only a variational thing)</p><p> </p><p>I'd use something like <code>initial_state = Statevector.from_label("00")</code> or at the very least stress that we are only referring to the ansatz with the initial parameters out of convenience, as we know that the state is the $|00\rangle$ (you already mentioned that fact but it can be helpful to bring that up here).</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Qiskit/qiskit-tutorials/pull/1382/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

Guillermo-Mijares-Vilarino, line 1: <p>I believe you can avoid this warning if you explicitly take the real part of the exact magnetization. As it is now, the <code>exact_magnetization</code> is a list of <code>complex</code> numbers with imaginary part 0.</p><p> </p><p>Also when I run the code as it is now I got the error <code>module 'scipy' has no attribute 'linalg'</code> . I had to add <code>import scipy.linalg</code> for it to work. Might be a problem with my version</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Qiskit/qiskit-tutorials/pull/1382/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

Guillermo-Mijares-Vilarino, line 1: <p>It might be nice to plot the energies from the <code>exact</code> function you defined previously to show that the exact solution should preserve the energy. With the current version of your code they are not used at all.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Qiskit/qiskit-tutorials/pull/1382/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='30'/>

Cryoris: <p>Yes, since this is a short tutorial I think we expect users to know (or look up) the basics. The PauliEvolutionGate has some more details and the link to a paper, I added that link </p>

Cryoris: <p>I intentionally used the ansatz, since that code will hold for any circuit and initial point. Giving the explicit state would only work for certain states and has some additional assumptions, so as it is it's more general :) </p><p> </p>

Cryoris: <p>good point, I added the <code>.real</code> ! Yeah scipy is a bit weird sometimes depending on the local installation... </p>

Cryoris: <p>good point :+1: </p>

Guillermo-Mijares-Vilarino: Yeah you have a point. The way I suggested in the previous comment would only work for basis states, so it'd be fine for the specific case of this tutorial but not in general 😅.

tutorials/algorithms/10_pvqd.ipynb
```diff
@@ -0,0 +1,584 @@
...
+    "%qiskit_copyright"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": []
+  }
```
ElePT, line 561: I am not sure if the suggestion is right, but basically I am trying to remove the last empty code cell of the notebook. Probably easier to do locally. Otherwise I think that the tutorial looks great!

```suggestion
  }
```

Cryoris: done!

Guillermo-Mijares-Vilarino:
It seems you are getting a timeout error now. You might want to change the following lines from `conf.py` to change the 180 seconds timeout. You can remove it entirely by changing the 180 to a -1

> cell_timeout = int(os.getenv('QISKIT_CELL_TIMEOUT', 180))
> nbsphinx_timeout = cell_timeout

Cryoris:
Do we still want to run these notebooks in a short amount of time? Because indeed the gradient descent cell does take some minutes to run. Alternatively we could reduce the number of timesteps, e.g. to 10 instead of 100, which might be fast enough for whatever the default is?

Guillermo-Mijares-Vilarino:
> Do we still want to run these notebooks in a short amount of time? Because indeed the gradient descent cell does take some minutes to run. Alternatively we could reduce the number of timesteps, e.g. to 10 instead of 100, which might be fast enough for whatever the default is?

Ideally yes. It's better to reduce the timesteps to make sure you don't surpass the 180 second limit. It's possible that your local machine takes a bit less than CI to run the cells.

Cryoris:
I reduced the number of timesteps to 10, locally that runs about 60s -- let's see if that worked 🙂

Cryoris:
@Guillermo-Mijares-Vilarino the checks seem to have passed, even though the status is not reported to GitHub 😄

PR #1383, opened 2022-12-09
Small fixes of 1_getting_started_with_qiskit.ipynb

filippotramonto:
### Summary
Small fixes of the tutorial 1_getting_started_with_qiskit:
- "`execute` function" --> "`run` method" in the text (in the code cells was already correct)
- "addition operator" --> "compose method" in a comment of a code cell
- there was some backtick too much, so the fix is: "``" --> "`" (for example, ``job`` --> `job`)

### Details and comments
see above


arne-kapell (approved):
only the described small fixes, nothing is changed code-wise

(I noticed the mistakes too, so I went looking for an already opened PR)

HuangJunye (approved):
LGTM. Thanks for the fix!

HuangJunye:
@arne-kapell Thanks for the review as well. Your efforts are appreciated!

PR #1385, opened 2022-12-22
Add primitives tutorial

tula3and:
### Summary

Added a tutorial related to Quantum Primitives.

Co-authored-by: Vishal Bajpe @mrvee-qC-bee 
Co-authored-by: Ikko Hamamura @ikkoham 

### Details and comments

A part of description and codes is based on [lab 1 notebook](https://github.com/qiskit-community/ibm-quantum-challenge-fall-22/blob/main/content/lab-1/lab1.ipynb) of the IBM Quantum Challenge 2022.


ikkoham:
Related to https://github.com/Qiskit/qiskit-tutorials/issues/1380

tutorials/primitives/01_introduction_to_primitives.ipynb
```diff
@@ -0,0 +1,936 @@
+{
```
javabster, line 1: <p>I like this diagram, but it feels a bit weird having the start box on the right hand side and then reading right to left. Can you invert the diagram so that the user is reading from left to right?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Qiskit/qiskit-tutorials/pull/1385/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='44'/>

mrvee-qC-bee: How does this look? @javabster @ikkoham FlippyflipQRT

mrvee-qC-bee: How does this look @javabster @ikkoham? 😁 FlippyflipQRT

woodsp-ibm:
I know you have the selection guide towards the end, but I wonder if it answers the question I see from most new users `how do I set the backend?` The way things work now is that a provider supplies primitives, like it supplied backends before. Not all providers/services have different backends that make such a selection meaningful in general. The runtime primitive you can choose. the reference and Aer not. The BackendEstimator/Sampler you can but these are really for legacy provider support and would not expected to be as performant at all as using a primitive from the provider, where it exists. I wonder if some section, more towards the front, stating this aspect would help.

Often the question of what primitive is addressed by what you want to do. If you are starting an experiment, where you would like an ideal outcome like you would get from using statevector simulator in the past, then choose the reference or Aer simulator primitives - care is needed since the defaults for ref is ideal but Aer needs to be configured that way. If you then want to see how sampling noise (shots) impacts things the former can be configured to do this and you can also consider the runtime simulator. Adding device noise to a simulator might be the next step. Aer sim you can do this and the runtime one too. The runtime sim you can use some of the noise mitigation techniques to see how they perform too ahead say finally running on a real device. So in my mind the selection can be driven this way - i.e. this thought process can be influencing your path down the above chart.

tutorials/primitives/01_introduction_to_primitives.ipynb
javabster: I think this looks better going from left to right but I agree with steve the the Yes/No is maybe a bit too aggressive with the colours here, perhaps something more neutral or not coloured box at all would be better

mrvee-qC-bee: How does this look? 😁 @javabster @woodsp-ibm 

Light Blue Modern Business Twitter Header (56)

mrvee-qC-bee:
I think this is a great suggestion @woodsp-ibm! Will try to formulate something along those lines so that the content flows down logically to the selections part at the end. @ikkoham @tula3and what do you guys think? I can share a draft along those lines to you guys!

ikkoham:
In my mind, the correspondence is as follows.

```
Reference in Terra: BasicAer
Aer primitives with `approximation=False`: PauliExpectation with Aer
Aer primitives with `approximation=True`:  AerPauliExpectation
BackendPrimitives: no error mitigation, but it can use external providers.
RuntimePrimitives: various error mitigation and suppression options, but provider is restricted (IBM devices including simulator)
```

How is BasicAer placed in your workflow?

tutorials/primitives/01_introduction_to_primitives.ipynb
ikkoham: Nice! Please push this to the PR and let's merge this PR.

Eric-Arellano:
Thanks for working on this! As shown by the emoji reactions, folks are definitely excited about this :) 

We're archiving this repository very soon and moving the tutorials over to qiskit-terra and qiskit-algorithms to build them directly in the documentation directly.

I believe this should be moved to qiskit-terra. Could you please reopen this PR over there?

PR #1449, opened 2023-05-04
Fix catch in grover tutorial for tera 0.24.0 compatibility

mtreinish:
### Summary

Qiskit Terra 0.24.0 had a bug that inadvertently changed the exception type returned when using the PhaseOracle class without having tweedledum installed locally. This commit changes the exception we catch in the grovers tutorial using the phase oracle to be an ImportError instead of the more specific MissingOptionalLibraryError exception. For terra 0.24.0 it is returning a ModuleNotFoundError instead of the correct MissingOptionalLibraryError, this will be fixed in 0.24.1, but in the meantime we can just use ImportError as it's the common parent class for both exception types and will be compatible with both.

### Details and comments

See: https://github.com/Qiskit/qiskit-terra/issues/10079 for the details of the qiskit-terra bug tracking the underlying issue.


Eric-Arellano (approved):
Maybe catch both MissingOptionalLibraryError and ImportError so that this is more future proof?

yoshida-ryuhei:
@mtreinish The updated tutorial is like [tutorial](https://qiskit.org/documentation/tutorials/algorithms/07_grover_examples.html) and it says `No module named 'tweedledum'`. This is not the best tutorial for me. Is this the intended one?
If not I will be happy to make a pull request to correct the tutorial.

PR #1451, opened 2023-05-10
Remove outdated docs template override & CSS files

Eric-Arellano:
HTML templates should not be overridden without a good reason. That messes up updates to qiskit_sphinx_theme and causes this docs build to diverge from qiskit-metapackage.

Likewise, this removes some unnecessary CSS files:

* `gallery.css` is for `sphinx-gallery`, which is not used by this project.
* `style.css` is leftover from when we used the Read The Docs theme. It's no longer relevant to qiskit_sphinx_theme.
* `custom.css` only has rules for `.toggle`, which does not exist anymore.

Finally, this activates jQuery because Sphinx 6 stopped activating it by default.


conf.py
```diff
@@ -64,10 +64,10 @@
     'sphinx.ext.mathjax',
     'sphinx.ext.extlinks',
     'nbsphinx',
+    "qiskit_sphinx_theme",
```
HuangJunye, line 67: Is this necessary now because the them ships with custom directives and other things?

HuangJunye (commented):
LGTM. Thanks for cleaning up the unnecessary css files. I have a small comment about the reason of adding `qiskit_sphinx_theme` to extensions. It's also not clear to me how this PR activates jQuery. How does that work?

conf.py
Eric-Arellano: In 1.11, we have qiskit_sphinx_theme automatically activate jQuery. But you have to put qiskit_sphinx_theme in the extensions list to have that work.

In 1.12+, it's good to also have qiskit_sphinx_theme in the extensions list because it activates the custom directives, Translations feature, and Previous Releases feature. So, we now expect all projects to have this config when they use our theme. I've finished updating every project to do that.

HuangJunye (approved):

conf.py
HuangJunye: Thanks for the detailed answer. That's very helpful!

PR #1455, opened 2023-05-16
Update note about pulse gate support

wshanks:
Pulse gate support is not tied to the open_pulse setting in `backend.configuration()`.

### Summary

Updates the note about pulse gate support so that it does not reference `open_pulse` which is a different setting.

### Details and comments

The `open_pulse` setting marks if a backend supports submitting pulse schedules as a job payload. IBM devices have deprecated submitting pulse schedules but still support pulse gates. There is no setting to check if pulse gates are supported (but currently they are supported on all IBM devices).


HuangJunye (approved):
Thanks for the update

PR #1459, opened 2023-05-18
Switch to using Tox

Eric-Arellano:
### Summary

Previously, there was no standardized way to build the docs locally. While Tox isn't perfect, it aligns us with the rest of the Qiskit ecosystem.

Switching to Tox is important here because it de-duplicates our requirements, which were specified both in requirements-dev.txt and again in azure-pipelines.yaml.

### Details and comments

This also fixes two Sphinx warnings due to bad config in `conf.py`.

I couldn't actually get NBSphinx to work on my mac due to import errors. But this is still some forward progress, so I put up the PR. CI at least will now use tox, and continue to produce a zip of the docs that CI users can inspect.


conf.py
```diff
@@ -12,34 +10,12 @@
...
-Sphinx documentation builder
-"""
-
-import os
-import sys
-# Set env flag so that we can doc functions that may otherwise not be loaded
-# see for example interactive visualizations in qiskit.visualization.
-os.environ['QISKIT_DOCS'] = 'TRUE'
-if sys.platform == 'darwin':
-    os.environ['QISKIT_IN_PARALLEL'] = 'TRUE'
```
Eric-Arellano, line 42: These env vars aren't used by anything. Bad copy and paste.

conf.py
```diff
@@ -69,21 +38,14 @@
...
-    "--InlineBackend.figure_formats={'png', 'pdf'}",
-    "--InlineBackend.rc={'figure.dpi': 96}",
-]
-
-nbsphinx_thumbnails = {
-}
-
+nbsphinx_widgets_path = ""
+html_sourcelink_suffix = ""
+nbsphinx_thumbnails = {"**": "_static/no_image.png"}
```
Eric-Arellano, line 48: I aligned this config with qiskit-metapackage, since that is what actually ends up being used in the final rendered docs.

conf.py
```diff
@@ -113,12 +75,7 @@
 numfig_format = {
     'table': 'Table %s'
 }
-# The language for content autogenerated by Sphinx. Refer to documentation
-# for a list of supported languages.
-#
-# This is also used if you do content translation via gettext catalogs.
-# Usually you set "language" from the command line for these cases.
-language = None
```
Eric-Arellano, line 121: Bad config in Sphinx 6+

tox.ini
```diff
@@ -0,0 +1,18 @@
+[tox]
+minversion = 3.15
+envlist = py310,py39,py38
+
+[testenv]
+no_package = true
```
Eric-Arellano, line 6: This is because we don't distribute any Python package in this repo. It's solely docs.

conf.py
HuangJunye: Great! This solve the thumbnail issue in the local build. In general, it's nice to see the same output in local build of the tutorials repo alone as the final rendered docs built by the metapackage.

conf.py
HuangJunye: Nice. CI is using Sphinx 7.0 instead of 5.3

tox.ini
```diff
@@ -0,0 +1,18 @@
...
+install_command = pip install -c{toxinidir}/constraints.txt -U {opts} {packages}
+deps =
+  -r{toxinidir}/requirements-dev.txt
+
+[testenv:docs]
+commands =
+  sphinx-build -j auto -b html {toxinidir} {toxinidir}/_build/html/
```
HuangJunye, line 13: Should we use `-W` flag as well or this repo is not ready for that yet?

HuangJunye (commented):
Thanks for working on this. It's great to reduce the duplication of dependencies in `azure-pipelines.yml` and making the local build closer to the final build by the metapackage.

Can you also update the instructions about how to build the documentation in readme to reflect the change? After that it should be ready for merging.

HuangJunye:
You mentioned that nbsphinx is not working on your mac. Does it mean that you can't make a successful local build using tox on your mac?

Eric-Arellano:
> Does it mean that you can't make a successful local build using tox on your mac?

Exactly. But I time-boxed trying to figure this out—there are a few other higher priorities I'm focused on. The main reason I'm cleaning up this repo is to unblock enabling `-W` in qiskit-metapackage, which is blocked by enabling `-W` in this repo.

Some forward progress is better than none. While I couldn't get local to work on my mac, CI at least still works.

PR #1466, opened 2023-06-02
Revert "Fix catch in grover tutorial for tera 0.24.0 compatibility (#1449)"

Eric-Arellano:
Now that Terra 0.24.1 has been released, we can revert this change. 

This doesn't actually impact what we deploy to https://qiskit.org/documentation/tutorials/algorithms/07_grover_examples.html, which depends on the metapackage repository: https://github.com/Qiskit/qiskit-metapackage/pull/1757. But it's good to have this notebook showing something more useful.


frankharkins:
Would it break anything to just remove the try/except blocks completely? They allow the notebook to "run" if tweedledum isn't installed, but it seems pointless as the notebook produces no output.

Eric-Arellano:
In general, this tutorial is a challenge because it means that we can't build docs on Apple Silicon, which is as of yesterday the sole type of Mac now being sold. 

Do we know how relevant this tutorial still is? Is it possible to rewrite it or delete it?

Eric-Arellano:
> Do you just need to build on M1 for local testing? If so, you might be able to use the raises-exception tag instead of the try/except blocks https://nbsphinx.readthedocs.io/en/0.9.2/allow-errors.html

Yeah, it's important that devs can build docs locally. It encourages them to check the quality of docs more. That was a project when I first joined to get `tox -e docs` working.

Rather than `raises-exception`, I think the best move would be improving the try/except to instead assert our specific conditions: if it's Python < 3.11 and not macOS, then eagerly error. Else, silently do nothing. That is because Tweedledum can't handle those situations. We started discussing in the Qiskit dev meeting this week what the long term plan for tweedledum is.

frankharkins:
Ok yeah that makes sense, this has more parts to it than I realised. Don't let this block merging the PR.

Issue #1405, opened 2022-06-21
Improve Advanced VQE usage tutorial if needed


y4izus:
Part of Qiskit/qiskit-tutorials#1401

PR #1471, opened 2023-06-09
Improving 04_vqd.ipynb

TannerTorrey3:
### Summary
Closes https://github.com/Qiskit/qiskit-tutorials/issues/1405

Replaced deprecated Classes with supported Classes
SumPauliOP -> SparsePauliOp
eigen_solvers.eigen_solver.Eigensolver -> eigensolvers.Eigensolver

### Details and comments

-Deprecated qiskit.opflow.primitive_ops.pauli_sum_op.PauliSumOp  Class replaced with qiskit.quantum_info.SparsePauliOp Class
-Deprecated qiskit.algorithms.eigen_solvers.eigen_solver.Eigensolver Class replaced with qiskit.algorithms.eigensolvers.Eigensolver Class


tutorials/algorithms/04_vqd.ipynb
```diff
@@ -163,15 +163,18 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 6,
+   "execution_count": 28,
    "metadata": {},
    "outputs": [],
    "source": [
     "from qiskit.algorithms.eigensolvers import VQD\n",
+    "from qiskit.algorithms.eigensolvers import NumPyEigensolver\n",
```
woodsp-ibm, line 171: This new import is not needed here right. Its imported below by the cell that uses the NumPyEigensolver

tutorials/algorithms/04_vqd.ipynb
```diff
@@ -163,15 +163,18 @@
...
+    "\n",
     "\n",
     "vqd = VQD(estimator, fidelity, ansatz, optimizer, k=k, betas=betas, callback=callback)\n",
     "result = vqd.compute_eigenvalues(operator = H2_op)\n",
-    "vqd_values = result.optimal_values"
+    "vqd_temp = result.eigenvalues\n",
+    "vqd_values = vqd_temp"
```
woodsp-ibm, line 177: There is no need for this to be changed at all, It was correct as it was before. Yes it may have this other field too but then,as you needed to do below, since its complex, you need to deal with that.

tutorials/algorithms/04_vqd.ipynb
```diff
@@ -253,19 +254,21 @@
...
+   "execution_count": 31,
    "metadata": {
     "pycharm": {
      "name": "#%%\n"
     }
    },
    "outputs": [],
    "source": [
+    "from qiskit.circuit import ParameterVector\n",
+    "from qiskit.quantum_info import SparsePauliOp\n",
```
woodsp-ibm, line 266: I am not sure why the ParameterVector import is needed at all. SparsePauliOp was imported above when `H2_op` was defined and should not be needed. The change needed here is just the removal of PauliSumOp import and its usage with H2_op, which you have done. The new NumPyEigensolver can take a SparsePauliOp (which is what H2_op is) directly and while it supports PauliSumOp, for legacy reasons, that is now deprecated hence the desire for this to be updated.

woodsp-ibm (changes requested):

tutorials/algorithms/04_vqd.ipynb
TannerTorrey3: My apologies for the wasted time, I attached the wrong .ipynb file initially. I will make remove these changes.

tutorials/algorithms/04_vqd.ipynb
```diff
@@ -169,9 +169,10 @@
    "source": [
     "from qiskit.algorithms.eigensolvers import VQD\n",
     "\n",
+    "\n",
     "vqd = VQD(estimator, fidelity, ansatz, optimizer, k=k, betas=betas, callback=callback)\n",
     "result = vqd.compute_eigenvalues(operator = H2_op)\n",
-    "vqd_values = result.optimal_values"
```
woodsp-ibm, line 174: I did not realize at first why you change that from `optimal_values` to `eigenvalues`. I would have thought that would have been the better field to use as was done before when this was run and checked in here. It seems however that there is an issue, which I raised here, Qiskit/qiskit-terra#10263 where the output from the field is not the values, but the points as can seen in the html which is published out there in the docs.

So this seems a reasonable workaround to have the energy values.

I am though concerned by the lowest energy value - that seems quite a way off (0.021) from where it should be, and evidently was before. The ground state computation for VQD should be doing the same as what VQE does and for H2, a very simple molecule, which is what the SparsePauliOp is the hamiltonian for it should get that answer. Since it has such good values for the 2 excited states, which depend on orthogonality from the ground state, it seems odd. I will note the ansatz shown above is for 3 qubits. Which while that's kinda odd the operator is only 2 qubits so it will get resized by VQD so I would think that confusing - I am not sure why it was done like that - I was looking at the ansatz since that can affect how well the the ground state can be determined.

TannerTorrey3: @woodsp-ibm  There is inherently some variation with the VQD values and the VQE's. I am only getting errors after 5 decimal places on my end. I can make the `optimal_values`  patch and then use the optimal_values instead of eigenvalues, removing the numpy.real. I believe the cast to real values from complex points is causing some loss or inaccuracy. Let me know if you agree with my course of action.

woodsp-ibm: If that first value is around 1.857 then it would be ok - at 1.837 (rounding it up) its a bit far where it should be was my main concern.

Taking the real part of a complex should not be a problem. 

For this tutorial to use optimal values a bug fix release would have to be done on Terra that included a fix in that regard. If you want to see about fixing the bug as per that issue it should be simple enough - but it maybe better for now to leave it as you have it with eigenvalues since that will work with the current code too.

woodsp-ibm: I copied the notebook from your fork and ran it and got `[-1.85727399 -1.24458395 -0.88272262]` which seems fine. I see some small variance between runs - this, I think, is because VQD, by default, chooses a random initial point. Maybe the outcome you committed was really unlucky in chosing an initial point from which it failed to converge well to the minimum. Not sure. I think if you commit something where the outcome is -1.857 something it would be better.

woodsp-ibm: It does seem to be the random initial point. If I add this in the VQD cell to seed the rng it uses
```
from qiskit.utils import algorithm_globals
algorithm_globals.random_seed = 220
```
then that value produces the 1.836 each time. Other seed values work eg 210. Though 50 I noticed fails on the excited states and they both end up incorrect and the same.

Bottom line I think its fine as it is. Maybe just commit one where the outcome is the right value. Or you can add that rngs seeding code so anyone running the notebook sees the same thing and does not randomly fail through a poorer initial point starting place.

Sorry, this turned out to be a bit more.... I thought the only issue with the notebook was the deprecated aspect!

woodsp-ibm: Its more to do with the initial point as I tried to explain above. I think the anstaz is fine - though arguably it could be confusing that its set to 3 qubits upfront when the operator is a 2 qubit operator. Since the ansatz is a template based circuit VQD can, and does, adjust the number of qubits to match. You can do this `print(result.optimal_circuits[0].decompose())` and print the circuit from the result to see this!

TannerTorrey3: Ok I made the anstaz change and it I seems to give me a similar error percentage. I will commit with two qubit anstaz with close values. Let me know if you prefer the three qubit I can revert the changes. 
With regards to your apology, I'm learning a lot so I actually enjoyed the added complexity. Sorry for my limited understanding and thank you for your patience.

woodsp-ibm: Changing the ansatz there, for number of qubits, will have no net bearing on the outcome. Internally VQD will try to adjust any ansatz to match the operator https://github.com/Qiskit/qiskit-terra/blob/86df15d3b9d80a74db91590af44af72a050371d2/qiskit/algorithms/eigensolvers/vqd.py#L180 This was done so you could pick say a TwoLocal ansatz and not need to know how many qubits the operator was - which if you specific a higher level problem you may not know at the time. Later the number of qubits can be, and is, set by VQD as it needs to match. My main comment there was more that it was showing a 3 qubit ansatz but it will end up setting it and using it as a 2 qubit one so its kinda misleading to show 3.

woodsp-ibm: Can you update this text there too, to say two instead of three

>You can set up, for example, a TwoLocal ansatz with three qubits, and choose SLSQP as the optimization method.

woodsp-ibm (approved):
@TannerTorrey3 Thanks for sticking with this and sorting it out. Updating it to address the deprecated usage turned out more than I expected!

@ElePT FYI and in case you have any comment.

TannerTorrey3:
@woodsp-ibm Thank you for the help! Can I email you some questions regarding a Quantum Computing club I'm founding? I promise to keep it short out of respect for your time. 

@ElePT Let me know if you would like additional assistance with optimal_values. I have some time constraints but would love the challenge.

woodsp-ibm:
> Can I email you some questions regarding a Quantum Computing club I'm founding? I promise to keep it short out of respect for your time.

I am not sure what sort of questions you have in mind. I am a maintainer for algorithms and applications so I look after technical matters in that area. If you looking to promote this club or perhaps gain ideas from others who have done similar I would look to Qiskit Slack - if you did not already join there is a link to join on the qiskit.org home page - scroll to the bottom and its under `Stay Connected` on the right hand side. There is even a `#meetups` channel there. If you feel its something technical, that I may be able to more directly answer for you, then sure.

ElePT:
@TannerTorrey3 `optimal_values` should be fixed by https://github.com/Qiskit/qiskit-terra/pull/10279 :)

divshacker:
fixes #1405

Eric-Arellano (approved):
Thank you for making Qiskit better!

PR #1475, opened 2023-06-16
Clarification of scope of optimal_num_iterations.

JGeipel:
### Summary

Quick change to Grover's Algorithm tutorial to clarify that `optimal_num_iterations` is designed for the uniform superposition and not general state preparations.

### Details and comments

Change was inspired by [this Quantum Computing StackExchange question](https://quantumcomputing.stackexchange.com/questions/32951/find-the-number-of-iterations-for-amplitude-amplification-to-get-the-correct-sta).


Eric-Arellano (approved):
Nice work! Thank you for making Qiskit better :)

PR #1478, opened 2023-06-21
Update README.md in preparation for archiving

1ucian0:


Eric-Arellano (commented):
Luciano, I think we should merge this this week. But I don't think we need this as detailed as tutorial level. It can be folder level:

- algorithms goes to qiskit-algorithms
- simulators goes to aer
- textbook was deleted
- everything else goes to qiskit-terra

Eric-Arellano (approved):

PR #1479, opened 2023-06-21
Remove the "Sample algorithms in Qiskit" section, as it is duplicated

1ucian0:
IQPE is the only entry in "Textbook algorithms" and it is duplicate with the IQPE in algorithms section. This PR removes the duplication.


Eric-Arellano (approved):
Thanks for cleaning this up.

PR #1480, opened 2023-06-25
fix missing backticks for plot_histogram()

soon-teh:
### Summary
The function plot_histogram() in the text should be `plot_histogram()`


Eric-Arellano (changes requested):
Hi, thanks for this improvement! We appreciate you making Qiskit better :)

It looks like this commit changes 1153 lines, when we want it to only change 1 line. In looking at the commit locally with Git, it looks like `^M` is being added to every single line. I'm not sure why, but possibly from your text editor? What text editor / IDE are you using?

To fix this, you can use `git revert 1feebb9` to revert your changes, and then have a new commit that only changes the 1 single line. Make sure before committing that the diff only has one line changed.

Eric-Arellano (dismissed):

Eric-Arellano (approved):

PR #1481, opened 2023-06-26
Update 02_gradients_framework.ipynb

pearcandy:
## Typo correction 

removal of "\\partial" in the second term on the right-hand side

### DIFF
--- a/tutorials/operators/02_gradients_framework.ipynb
+++ b/tutorials/operators/02_gradients_framework.ipynb
@@ -548,7 +548,7 @@
     "\n",
     "Unlike the other methods, finite difference gradients are numerical estimations rather than analytical values.\n",
     "This implementation employs a central difference approach with $\\epsilon \\ll 1$\n",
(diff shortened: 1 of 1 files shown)


pearcandy:
I am new to quantum computation and was just learning about gradients. At the time, I felt that I did not need the partial symbol on the right side in the finite difference formula. If I am wrong, please ignore my pull request.

Eric-Arellano:
Okay, thanks. I'm going to assume that the original author and code reviewers were correct in having it, then.

In the future, when making contributions, it is helpful to put in the PR description what the motivation for the change is. If you're not sure whether a change is good or not, either don't open it or ask for feedback, e.g. by opening a GitHub issue or using the project's Slack.

Cheers in your quantum computing journey!

pearcandy:
Thanks for your advice. Next time, I will try to discuss on Slack projects.

In this issue, I think the symbol for partial on the right-hand side in the equation, i.e., \partial \langle \psi(\theta..., does not make sense mathematically.  This is because the equation implies rewriting the continuous derivative with respect to ε into a discrete numerical derivative. Nevertheless, the sign of the partial derivative remains on the right-hand side, which is obviously not correct. It is clear from a comparison with the formula for Parameter Shift Gradients discussed above in this chapter.

PR #1467, opened 2023-06-03
Improves 03_vqe_simulation_with_noise

divshacker:
### Summary
This PR will Improve [03_vqe_simulation_with_noise](https://qiskit.org/documentation/tutorials/algorithms/03_vqe_simulation_with_noise.html)

### Details and comments
fixes #1404 
Part of #1401

PR #1445, opened 2023-04-27
Improves An introduction to algorithms

divshacker:
### Summary
This PR will improve the tutorial [01_algorithms_introduction](https://qiskit.org/documentation/tutorials/algorithms/01_algorithms_introduction.html)

### Details and comments
fixes #1402

PR #1487, opened 2023-07-05
Aer tutorials broken links

1ucian0:
This PR updates the Aer links to point to the right places. Thanks Voica for the heads-up!

### DIFF
--- a/tutorials/algorithms/01_algorithms_introduction.ipynb
+++ b/tutorials/algorithms/01_algorithms_introduction.ipynb
@@ -97,7 +97,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "This estimator uses an exact, statevector simulation to evaluate the expectation values. We can also use a shot-based and noisy simulators or real backends instead. For more information of the simulators you can check out [Qiskit Aer](https://qiskit.org/documentation/apidoc/aer_primitives.html) and for the actual hardware [Qiskit IBM Runtime](https://qiskit.org/documentation/partners/qiskit_ibm_runtime/).\n",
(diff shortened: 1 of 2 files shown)


tutorials/algorithms/01_algorithms_introduction.ipynb
```diff
@@ -97,7 +97,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "This estimator uses an exact, statevector simulation to evaluate the expectation values. We can also use a shot-based and noisy simulators or real backends instead. For more information of the simulators you can check out [Qiskit Aer](https://qiskit.org/documentation/apidoc/aer_primitives.html) and for the actual hardware [Qiskit IBM Runtime](https://qiskit.org/documentation/partners/qiskit_ibm_runtime/).\n",
+    "This estimator uses an exact, statevector simulation to evaluate the expectation values. We can also use a shot-based and noisy simulators or real backends instead. For more information of the simulators you can check out [Qiskit Aer](https://qiskit.org/ecosystem/aer/apidocs/aer_noise.html) and for the actual hardware [Qiskit IBM Runtime](https://qiskit.org/documentation/partners/qiskit_ibm_runtime/).\n",
```
woodsp-ibm, line 100: This was linked to the Aer primitives before since that is the unit which algorithms consume - the runtime link has tutorials there specifically but that seemed to be the "best" link for Aer. Yes it mentions alternative modes, like adding noise model, but thats more for an overview here (tutorial 3 covered this more) here I think it should be to the primitives.

divshacker:
I have updated the broken links already in my PRs #1445 #1467 which improves these tutorials. These PRs are part of QAMP project, that's why they are not yet merged. They all will get merged next week.

tutorials/algorithms/01_algorithms_introduction.ipynb
ElePT: I think this might be due to a misplaced copy-paste:

```suggestion
    "This estimator uses an exact, statevector simulation to evaluate the expectation values. We can also use a shot-based and noisy simulators or real backends instead. For more information of the simulators you can check out [Qiskit Aer](https://qiskit.org/ecosystem/aer/apidocs/aer_primitives.html) and for the actual hardware [Qiskit IBM Runtime](https://qiskit.org/documentation/partners/qiskit_ibm_runtime/).\n",
```

